PR #24, opened 2016-12-16
Upload doc only option

craig-bishop:
I really love the capability to easily upload docs to gh-pages. However, I am working on private repositories and I only want to upload docs, not publish, tag, etc.

So I added a simple switch to only perform the doc upload.  I refactored the doc upload code to a separate function and added a short-circuit.


sunng87:
hi @craig-bishop , sorry to be sooo late to response. Thank you for the patch. But in my opinion this function might be slightly out of cargo-release's scope. If you want to upload doc only, you can use tools like [gh-import](https://pypi.python.org/pypi/ghp-import), which is quite handy for this task. But adding another option to cargo-release may add extra complexity to most typical users. So I won't merge this into mainline. Thank you again for your nice patch, includes docs update.

PR #32, opened 2017-04-27
Pre-release Hook

### LINKED ISSUE #31: pre-release: generate changelog
Capturing the original discussion from https://github.com/nix-rust/nix/issues/595.

As one of the last steps in a release process, there is often the need of semi-automatically generating a changelog.

I'd be better personally happy in having some field to specify a custom binary/script, and some calling convention to follow (e.g. to know about previous and new version, etc.). That way I could extend the process with further logic without making cargo-release too opinionated, and even introduce some interactive action in the middle for a final cosmetic pass.

sunng87:
Trying to resolve #31 

Added a pre-release-hook that allow user to assign a script/command that to be executed before committing the version change. Files modified in this stage will be committed in next step.


src/main.rs
```diff
@@ -137,6 +139,11 @@ fn execute(args: &ArgMatches) -> Result<i32, error::FatalError> {
             try!(replace::do_replace_versions(pre_rel_rep, &new_version_string, dry_run));
         }
 
+        if let Some(pre_rel_hook) = pre_release_hook {
+            println!("{}", Green.paint(format!("Calling pre-release hook: {} {} {}", pre_rel_hook, new_version_string, &prev_version_string)));
+            try!(cmd::call(vec![pre_rel_hook, &new_version_string, &prev_version_string], dry_run));
```
lucab: Rule of thumbs that I learned the hard way doing CLI binary interfaces:
 * Try to avoid having more than one positional arguments if the order matters, preferring named arguments (see `ln` for a bad example)
 * Don't use positional arguments for optional things
 * Try to use reasonably sized long-named arguments
 * If you plan to grow the interface but still don't know how much or in which direction, use (prefix-namespace) environment flags

In short: I would prefer those to be `hook [--old OLDVER] NEWVER`, or both moved to named (or environmental) arguments .

sunng87: @lucab , I world prefer environmental arguments. Thank you for these great suggestions. That's pretty helpful to me.

Issue #30, opened 2017-04-16
Issue when using inner crate


antoyo:
Hi.

When specifying a dependency with `path`, it is annoying to use `cargo release` because of the added suffix `-pre`.
It is an issue because you want to specify a version like `"^1.2.3"` which is incompatible with `"1.2.4-pre"`.

Currently, I use the option `--no-dev-version`, but I need to manually bump the version.

I would like to be able to have automatic version bump, but without the `-pre` suffix.

Is this possible to do so?

Thanks!

antoyo:
@sunng87 Any update on this issue?
I can help to fix it, if needed.

sunng87:
Hi @antoyo , I haven't got time for this. Currently if you do auto bump to `1.2.4`, the next time you call `cargo release` it will be updated to `1.2.5`. ([code](https://github.com/sunng87/cargo-release/blob/master/src/version.rs#L26))

So if you use `--no-dev-version`, we will automatically bump version in next `cargo release`. I guess that will work for you.

bluss:
@antoyo Is the inner crate published or not? If it's not published, the resolution is to drop the version key entirely in toml. Otherwise I only have the not using dev version fix..

antoyo:
They are published.
Could there be an option to bump versions without the suffix, please?
Thanks.

epage:
So I'm looking to see how this might be related to #77 which I'm working towards.

#77 is about the pre-release version bump.  This seems focused on the post-release version bump.  

Is there a reason the `no-dev-version` flag doesn't work for your use case?  It will make it so there is no post-release version bump.

That doesn't help if you want to do a post-release version bump without a suffix.  This effectively moves the pre-release version bump to post-release.  Is there a reason a pre-release bump is insufficient? Currently, when `cargo-release` does the pre-release bump, it assumes it needs to increment the update field if there is no suffix.  We'd need to set up flags and shuffle things around to support this workflow.

epage:
I believe this was resolved as part of #295.  If this specific case was not, feel free to open.  If people find a new case, feel free to create an issue.  In either case, please include the output with trace-level logging so we can get an idea of where things are going wrong

PR #33, opened 2017-05-01
Change default dev extension to alpha.0

sunng87:


lucab:
What's the rationale for this? As a datapoint, I typically use `-dev`, `-git` or `-pre` during devel cycle and only move to `-alpha.0` when I actually want to push a pre-release to crates.io.

sunng87:
according to https://github.com/sunng87/cargo-release/issues/30, cargo has issue to recognize `pre` extension as a valid version.

bluss:
If I'm reading it right, #30 is about that "^1.2.3" does not match a later version with prerelease tag, which is the same regardless of what the prerelease tag is (alpha, pre, dev whatever). This is by design in cargo.

PR #38, opened 2017-08-28
Allow specifying the tag prefix in Cargo.toml

### LINKED ISSUE #37: Tag prefix cannot be set via `[package.metadata.release]`
The tag prefix can be set by passing `--tag-prefix <prefix>`, but the same doesn't seem to be possible by setting a `[package.metadata.release]` option in `Cargo.toml`. At least that feature doesn't seem to be documented in the README (unless I missed it).

I tried setting the `tag-prefix` and `prefix` options, but both resulted in an error message.

romac:
This will only override the default tag prefix if `--tag-prefix` is not specified.

This should fix #37.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,8 @@ store these options. Available keys:
   version.
 * `tag-message`: string, a message template for tag. Available
   variables: `{{version}}`, `{{prefix}}` (the tag prefix)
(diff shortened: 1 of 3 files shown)


sunng87:
Thank you, @romac

romac:
No problem!

PR #39, opened 2017-12-12
make publish crate optional

ChriFo:
The official manifest has a [key](http://doc.crates.io/manifest.html#the-publish--field-optional) for disable publishing. 

Currently `cargo release` will fail when `cargo publish` is executed and **publish = false**.

With my PR no extra option is added!


sunng87:
Nice, thank you @ChriFo

Issue #34, opened 2017-05-22
IOError when rewriting version in `Cargo.lock` when project is inside a workspace


HellButcher:
Hi,
I have a workspace layout like this:
```
workspace_root/
+- Cargo.toml
|    [workspace]
|    members = [ "project", ...]
+- Cargo.lock
+- project/
|  +- .git/
|  +- Cargo.toml
|  |    [package]
|  |    name = "project"
|  |    version = "0.1.0"
|  |    ...
|  +- src/
+ ...
```

when I run `cargo release --level patch` inside `project/` (for example),
I get something like this:
```
$ cargo release --level patch
Release version 0.1.1 ? [y/N] y
Update to version 0.1.1 and commit
Fatal: IOError
```

I have looked into the source and found out, that `config::rewrite_cargo_version` tries to read the `Cargo.lock` file from the current directory.
As a workaround, I have copyed the `Cargo.lock` file from the workspace to the project, and the cargo-release worked.

sunng87:
Good catch. This tool wasn't well tested in workspace scenario. I will look into this issue soon.

avkonst:
I have got the same in non workspace scenario

avkonst:
It is the behavior on windows

sunng87:
@avkonst could you please show me a tree view of your project layout? And what's the version of cargo-release you were using?

avkonst:
Cargo.toml:
```
[package]
name = "adapter-node"
version = "0.2.8-alpha.0"
authors = ["konstaan"]
publish = false

[dependencies]
clap = "2.27.1"
error-chain = "0.11.0"
futures = "0.1.14"
hyper = "0.11.2"
tokio-core = "0.1"
serde = "1.0"
serde_derive = "1.0"
serde_json = "1.0"
chrono = { version = "0.4", features = ["serde", "rustc-serialize"] }
openssl-probe = "0.1.2"

[dev-dependencies.error-chain-for-dumb-ides]
version = "0"

```

Version:
```
cargo-release.exe-release 0.9.0-beta.6
```

Layout:
```
./.git
./.gitignore
./Cargo.lock
./Cargo.lock.work
./Cargo.toml
./Readme.md
./rust-exp.iml
./src
./src/main.rs
```

avkonst:
shall I report a separate issue for this?

sunng87:
No. I've got the issue. Let me check the code and find a Windows box to reproduce.

epage:
[...]

PR #40, opened 2017-12-22
Check if Cargo.lock exists before try to adjust version upgrade

ChriFo:
`cargo release` tries to update the version in the file **Cargo.lock**.
But **Cargo.lock** must not be present! 

From the official [guilde](http://doc.crates.io/guide.html#cargo.toml-vs-cargo.lock):
> If you’re building a library that other projects will depend on, put Cargo.lock in your .gitignore. If you’re building an executable like a command-line tool or an application, check Cargo.lock into git.

Currently if **Cargo.lock** is not present, e.g., after `git clone`, an error appears:
```shell
Update to version 0.14.3 and commit
Fatal: IOError
```
With this PR **Cargo.lock** will be touched only if present!

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -132,7 +132,7 @@ pub fn rewrite_cargo_version(version: &str) -> Result<(), FatalError> {
     }
     try!(fs::rename("Cargo.toml.work", "Cargo.toml"));
 
-    {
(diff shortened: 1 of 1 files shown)


ChriFo:
I think this PR also resolve #34

sunng87:
Thank you!

PR #42, opened 2018-01-25
Fix version update in Cargo.lock

ChriFo:
This should fix the version upgrade in the _Cargo.toml_ for the released crate.


sunng87:
Thank you for the quick fix!

PR #48, opened 2018-02-09
fix: Drop Cargo.lock before renaming it

Marwes:
Lets it work on Windows


sunng87:
Thanks! Is this to fix IOError on Windows?

Marwes:
It fixes a IOError on windows at least since it is not possible to modify a file while it is being read from.

sunng87:
Nice, thank you @Marwes

sunng87:
Nice, thank you @Marwes

sunng87:
Nice, thank you @Marwes

PR #53, opened 2018-03-19
pre-publish-hook

dalance:
I want to `cargo publish` for inner crate like the following directory structure.

```
inner-crate/
inner-crate/Cargo.toml
Cargo.toml
```

If `pre-publish-hook` is supported, I can write like below.

```
pre-publish-hook           = ["sh", "-c", "if [ \"$DRY_RUN\" = \"false\" ]; then cd inner-crate; cargo publish; else echo \"cd inner-crate; cargo publish;\"; fi"]
pre-release-replacements   = [{ file = "inner-crate/Cargo.toml", search = "version = \"[a-z0-9\\.-]+\"", replace = "version = \"{{version}}\"" }]
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -21,6 +21,7 @@ pub static PRE_RELEASE_COMMIT_MESSAGE: &'static str = "pre-release-commit-messag
 pub static PRO_RELEASE_COMMIT_MESSAGE: &'static str = "pro-release-commit-message";
 pub static PRE_RELEASE_REPLACEMENTS: &'static str = "pre-release-replacements";
 pub static PRE_RELEASE_HOOK: &'static str = "pre-release-hook";
+pub static PRE_PUBLISH_HOOK: &'static str = "pre-publish-hook";
 pub static TAG_MESSAGE: &'static str = "tag-message";
 pub static TAG_PREFIX: &'static str = "tag-prefix";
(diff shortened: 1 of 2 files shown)


sunng87:
Hi @dalance , will `pre-release-hook` work for you in this case?

dalance:
Hi, @sunng87 
Thank you for reply.

I think `pre-release-hook` can't use in this case.
I want to execute `cargo publish` of inner crate after `pre-release-replacements` and `git commit` of version bump,
but `pre-release-hook` is executed before `pre-release-replacements` like below.

```
pre-release-hook
pre-release-replacements
git commit
<- I want to insert `cargo publish` of inner crate in this point.
cargo publish
```

If `pre-release-hook` is moved after `pre-release-replacements`, I can use it.
In the case, I'll add `--allow-dirty` to `cargo publish` of inner crate because it is before `git commit`.

sunng87:
Actually I think the things you done in `pre-release-replacements` and `pre-publish-hook` can be replaced by a simple shell script, that calls `cargo release` in both inner crate and outer one, right?

I'm afraid your current usage made it difficult to understand and maintain. Perhaps we can improve multi-crate project support for cargo-release? It's on my todo.

dalance:
I understand your afraid.
I'll wait for multi-crate support.
Thank you.

PR #68, opened 2018-07-05
Add the `{{prev_version}}` variable

jonas-schievink:
It's replaced by the crate version before `cargo-release` changes it in any way (ie. the version read straight from the `Cargo.toml`).

This can be used, for example, to automatically create links to the diffs between two versions in a changelog.


sunng87:
Good addition! Thanks! I'll create a release soon.

PR #70, opened 2018-08-17
config: add a "skip-publish" option

lucab:
This introduces a new configuration option "skip-publish", disabled
by default. It is similar to the "publish" parameter in Cargo.toml,
but can be toggled on invidual `cargo release` runs without changing
the manifest.


lucab:
The usecase for this is when a release is being prepared on a machine (or by a user) which does not have token/permissions to directly push to crates.io.

PR #71, opened 2018-08-17
config: undeprecated metadata from cargo manifest

lucab:
This undeprecates the source of configuration from cargo manifest
metadata section. It also massages the code-flow to avoid deeply nested
conditionals.

Ref: https://github.com/sunng87/cargo-release/pull/54#issuecomment-413849959

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -10,7 +10,6 @@ use semver::Version;
 use toml::value::Table;
 use toml::{self, Value};
 
-use shell;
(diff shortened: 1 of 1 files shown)


sunng87:
Thanks! I will create a release soon

lucab:
@sunng87 great, thank you! Let me know once it is published on crates.io.

sunng87:
0.10.4 is pushed.

PR #74, opened 2018-12-12
Revamp replacements

LegNeato:
This revamps replacements to make them more consistent.

Changes:

* All strings can have placeholders.
* Crate name placeholder for strings is available as `{{crate_name}}`.
* *Breaking Change*: Instead of subdir, tag prefix uses the crate name from `Cargo.toml`. In practice the crate name should usually be the same as the subdir anyway.
* *Breaking Change*: `{{version}}` in `pro_release_commit_msg` now refers to the
version that was released. To refer to the version that will be in git after
`cargo-make` runs, use `{{next_version}}`.

Let me know if this should be split out or if that last change is undesired. It felt inconsistent that `{{version}}` changed depending on what config key you put it in, and the way it was prevented you from using the just released version in `pro_release_commit_msg`.


LegNeato:
I think now that we are doing this the whole mental model can be simplified.  For example, why do we need `tag-prefix` when we could just specify `tag` like `myprefix-{{version}}`. I think the prefix thing was likely there because you didn't do replacements on everything?

sunng87:
Thanks! The idea sounds pretty good to me!

Yes, by default the tag was just version numbers. To add some customization we supported tag prefix for some users who want `release-v1.1` in case they have other tags. And using these global variables we can more consistent and flexible. 

I will walk through the details in next few days.

src/main.rs
```diff
@@ -287,11 +296,11 @@ fn execute(args: &ReleaseOpt) -> Result<i32, error::FatalError> {
             .push(Identifier::AlphaNumeric(dev_version_ext.to_owned()));
         shell::log_info(&format!("Starting next development iteration {}", version));
         let updated_version_string = version.to_string();
+        replacements.insert("{{new_version}}", updated_version_string.clone());
```
sunng87: README says it's `{{next_version}}` here

sunng87:
ping

src/main.rs
LegNeato: Changed to `{{next_version}}` everywhere.

src/main.rs
```diff
@@ -249,28 +259,27 @@ fn execute(args: &ReleaseOpt) -> Result<i32, error::FatalError> {
...
         .clone()
         .or_else(|| {
             config::get_release_config(release_config.as_ref(), config::TAG_PREFIX)
                 .and_then(|f| f.as_str())
                 .map(|f| f.to_string())
+                .map(|f| replace_in(&f, &replacements))
         })
-        .or_else(|| rel_path.as_ref().map(|t| format!("{}-", t)));
-
+        .or_else(|| Some(format!("{}-", crate_name)));
```
jonas-schievink: I'm pretty sure this caused https://github.com/sunng87/cargo-release/issues/81

PR #75, opened 2018-12-17
Adding feature-flags to release configuration

spacekookie:
This PR adds `features` and `all_features` as both flags to the CLI
as well as the `release.toml` configuration. This change allows you
to tell `cargo-release` to verify the package with certain features
enabled, in order to avoid needing to use `no-verify` for crates
that depend on certain features being enabled (or simply any
feature) before they can be uploaded.

This change is dependent on rust-lang/cargo#6453, which adds this
feature to the base `cargo publish` command.

The motivation behind this PR is the same as the above mentioned
cargo PR: reducing the amount of packaging errors that can happen
due to people not being able to verify their code because it has
some feature requirements for compilation, ultimately resulting in
a more healthy package ecosystem.


sunng87:
LGTM. Could you please also update README for new options?

spacekookie:
Hey there, apologies for the long wait!

I added the docs to `README.md` (I hope this is sufficient?) as well as resolving the merge conflict from master

sunng87:
Thanks!

PR #76, opened 2019-01-12
Reorder pre-release hook

antifuchs:
This change moves the pre-release hook so that it gets called right
before `git commit` for the release version happens.

This should ensure that if a repo sets a release hook to validate
release readiness, it can validate all the file replacements that
cargo-release has made.


sunng87:
Sounds good to me. Thanks for the patch.

PR #79, opened 2019-02-14
Code cleanup

epage:
Found these as I've worked on switching config handling to serde.


epage:
Finished rebasing this.

PR #88, opened 2019-03-03
 fix(cargo): Read CARGO on user's system 

epage:
`env!` reads the variable at compile time but this is a variable we
should be reading on the end-user's system, not the system
`cargo-release` was built on.

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::fs::{self, File};
 use std::io;
 use std::io::prelude::*;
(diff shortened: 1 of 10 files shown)


epage:
This also includes changes related to separating roles.  The bug was found while in the middle of it and I was being lazy and didn't want to deal with all of the conflicts from separating out that one commit.

I've not yet reached my end-goal with these changes; I felt I changed too much to keep progressing (don't want reviewing this to be too hard).

sunng87:
Yes. Thank you for keep pr relatively small and adding tests. That's very helpful!

src/cargo.rs
```diff
@@ -1,22 +1,203 @@
...
+use semver::Version;
+use toml::Value;
+
 use cmd::call;
 use error::FatalError;
 use Features;
 
-pub fn publish(dry_run: bool, features: Features) -> Result<bool, FatalError> {
+pub fn publish(dry_run: bool, manifest_path: &Path, features: Features) -> Result<bool, FatalError> {
+    let cargo = env::var("CARGO")?;
```
sunng87: I suggest to fallback to default cargo binary on PATH when `CARGO` is not defined.

PR #89, opened 2019-03-03
(fix) fallback to default cargo on PATH

sunng87:
This can be useful when testing `cargo-release` without cargo context.

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -13,8 +13,12 @@ use cmd::call;
 use error::FatalError;
 use Features;
 
+fn cargo() -> String {
+    env::var("CARGO").unwrap_or("cargo".to_owned())
+}
+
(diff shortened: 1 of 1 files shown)


src/cargo.rs
```diff
@@ -13,8 +13,12 @@ use cmd::call;
 use error::FatalError;
 use Features;
 
+fn cargo() -> String {
+    env::var("CARGO").unwrap_or("cargo".to_owned())
```
epage: FYI if you run clippy, it'll complain about a non-trivial `unwrap_or` and suggest the use of `unwrap_or_else`.

Thoughts on adding clippy to Travis?  Feel free to copy as much or as little of [my suggested travis workflow](https://github.com/crate-ci/escargot/blob/master/.travis.yml).

sunng87:
@epage Thanks! I will fix clippy issues in next patch and add CI check.

PR #91, opened 2019-03-07
Switch away from regex-based file editors

epage:
It just so happened that the non-regex parser I used for the lock file also works in workspaces


sunng87:
Thanks!

Issue #63, opened 2018-04-27
Integrate with cargo library


sunng87:
Cargo library provides a few apis for extension development. We will move onto that.

PR #98, opened 2019-03-12
Port to cargo_metadata where possible

epage:
For determining the default tag, I did not switch us from finding the git root to finding the workspace root.  I figured that if the crate wasn't in the git root, regardless of the use of workspaces, we probably want to scope the tag.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -100,7 +100,7 @@ dependencies = [
  "predicates 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "quick-error 1.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
- "semver 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


epage:
1. `cargo metadata` doesn't expose `publish`, that is weird
2. `cargo_metadata` doesn't expose the `metadata` table.  I understand (its freeform, what API to provide?).

epage:
nm, it does support metadata.  Somehow I missed that when going through the docs

https://docs.rs/cargo_metadata/0.7.1/cargo_metadata/struct.Package.html#structfield.metadata

epage:
1. Did a rebase
2. Fixed an `expect` failing when run in a workspace root without a package (`pure_ws` test condition)

sunng87:
Perhaps we can submit an issue for cargo_metadata to support `publish`?

Just knew about cargo_metadata is a great to for cargo ecosystem, I will close #63 because cargo_metadata is a better approach.

epage:
> Perhaps we can submit an issue for cargo_metadata to support publish?

iirc `cargo metadata` (which `cargo_metadata` wraps) doesn't expose `publish`, so we'd first need the issue resolved in `cargo metadata`.

But yes, it'd be nice if they exposed it.

PR #101, opened 2019-03-30
Update workspace dependents on release

### LINKED ISSUE #77: Verify / update corresponding version for path dependencies
In a workspace, I need to define my dependencies as `dep = { path = "../dep", version = "0.1.0" }` but I forget (until `cargo publish`) to update the versions for my path dependencies (and it is really annoying).

I think building this into `cargo-release` would be a big help.

I view this as a subset of the work for #66.  While that is for releasing an entire workspace, this will help with releasing parts of a workspace (and is needed for the entire workspace).

Configuration
- What to do on path dependency
  - verify
  - bump to latest (has to check crates.io because local path dependency has probably been bumped post-publish)
  - bump to semver compatible version

epage:
Fixes #77 

This includes switching to CLI arguments using `enum` where possible.  The downside to `arg_enum!` is that `--help` will show the names in uppercase and accepts them case insensitively which is uncommon in CLIs.   We probably want to eventually roll our own set of `impl`s rather than relying on `arg_enum!`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -163,8 +163,17 @@ pub fn resolve_config(manifest_path: &Path) -> Result<Option<Config>, FatalError
     Ok(None)
 }
 
-#[test]
-fn test_release_config() {
-    let release_config = resolve_config(Path::new("Cargo.toml")).unwrap().unwrap();
-    assert!(release_config.sign_commit);
+#[cfg(test)]
+mod test{
+    use super::*;
+
+    mod resolve_config {
+        use super::*;
+
+        #[test]
+        fn doesnt_panic() {
+            let release_config = resolve_config(Path::new("Cargo.toml")).unwrap().unwrap();
+            assert!(release_config.sign_commit);
+        }
+    }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -31,6 +31,7 @@ pub struct Config {
     pub disable_tag: bool,
     pub enable_features: Vec<String>,
     pub enable_all_features: bool,
+    pub dependent_version: DependentVersion,
 }
 
 impl Default for Config {
@@ -54,6 +55,7 @@ impl Default for Config {
             disable_tag: false,
             enable_features: vec![],
             enable_all_features: false,
+            dependent_version: Default::default(),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/config.rs
```diff
@@ -54,6 +55,7 @@ impl Default for Config {
             disable_tag: false,
             enable_features: vec![],
             enable_all_features: false,
+            dependent_version: Default::default(),
```
epage: Not a fan of this config flag name but couldn't think of a better one.

src/main.rs
```diff
@@ -363,6 +422,10 @@ struct ReleaseOpt {
     /// Do not create git tag
     skip_tag: bool,
 
+    #[structopt(long = "dependent-version", raw(possible_values = "&config::DependentVersion::variants()", case_insensitive = "true"))]
```
epage: Not a fan of this CLI flag name but couldn't come up with a better one

src/version.rs
```diff
@@ -1,67 +1,81 @@
...
-        },
-        None => {
-            if version.is_prerelease() {
-                version.pre.clear();
-                need_commit = true;
+            BumpLevel::Release => {
+                if version.is_prerelease() {
+                    version.pre.clear();
+                    need_commit = true;
+                }
```
sunng87: When using default `patch` level on pre-release it works exact this way. I suggest not to add another `Release` level.

sunng87:
Sorry for late review. The `dependentVersion` behaviour looks good to me and I agree it needs a better and simple name to understand.

src/version.rs
epage: I was maintaining the existing default which has different logic than patch.

Is your preference just to make `patch` the default instead?

src/config.rs
epage: >  I agree it needs a better and simple name to understand.

Should we block merging based on the name being less than desirable?  If so, how do we overcome this?

sunng87: Never mind. We can keep using this name in beta versions until we come up with a better one.

src/version.rs
sunng87: Checked again. The default behaviour is for the case that when release failed, for example, an error on `cargo publish`, the version is already bumped and we have a way to rerun `cargo-release`. Otherwise, it makes more sense to use `patch` by default. 

To fix the rerun issue, also mentioned in #103 , I am going to add a transaction log file like maven release, so we can always rerun the failed process.

PR #105, opened 2019-04-25
Adds doc on `disable-publish` flag

laysakura:
This flag is effective but just not documented.

Useful for projects which does `cargo publish` in CI.

Example here: https://github.com/laysakura/fid-rs/blob/v0.1.0/.travis.yml#L57-L62

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ Available configuration keys are:
 * `push-remote`: string, default git remote to push
 * `disable-push`: bool, don't do git push
 * `disable-tag`: bool, don't do git tag
-* `disable-publish`: bool, don't do cargo publish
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -178,6 +178,7 @@ Available configuration keys are:
 * `push-remote`: string, default git remote to push
 * `disable-push`: bool, don't do git push
 * `disable-tag`: bool, don't do git tag
+* `disable-publish`: bool, don't do cargo publish
```
epage: I think it'd be help to also call out here that we also respect the Cargo.toml publish flag.

Maybe
```md
* `disable-publish`: bool, don't do cargo publish right now, see [manifest `publish` field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish--field-optional) to permanently disable publish.
```

epage:
Thanks!

Issue #102, opened 2019-04-04
Cleanup code base?


epage:
My thought is to
- Run rustfmt and add it to the CI
- Fix clippy issues and add it to the CI
- Switch to the 2018 edition.

Creating this issue to first make sure this is all ok.  Since the work is pretty rote, I won't bother waiting on a review once I get the go ahead.  This will be disruptive to other PRs, so I'll do it when I am not aware of any PRs in the queue.

epage:
@sunng87 if you are ok with this, then timing will be important since it'll be disruptive enough that we won't want any active PRs when making the change over.

sunng87:
Thanks @epage , I'm OK with these improvements. Sorry for late response because I have been really busy with both my work and life. I will try best to catch up and respond your PRs in time.

epage:
No problems.  I've been busy as well since I initially did the dependency-version PR.

PR #108, opened 2019-04-27
Migrate to 2018 Edition

epage:
Finish up #102

### DIFF
--- a/src/version.rs
+++ b/src/version.rs
@@ -1,4 +1,4 @@
-use error::FatalError;
+use crate::error::FatalError;
 use semver::{Identifier, Version};
 use semver_parser;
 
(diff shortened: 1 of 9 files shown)


src/version.rs
```diff
@@ -297,15 +297,15 @@ mod display {
...
-        fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
             if self.0.predicates.is_empty() {
-                try!(write!(fmt, "*"));
+                r#try!(write!(fmt, "*"));
             } else {
                 for (i, ref pred) in self.0.predicates.iter().enumerate() {
                     if i == 0 {
-                        try!(write!(fmt, "{}", DisplayPredicate(pred)));
+                        r#try!(write!(fmt, "{}", DisplayPredicate(pred)));
```
sunng87: Didn't know about this? What's the difference between `r#try` and `?` ?

sunng87:
Failed on `cargo fmt --check`

src/version.rs
epage: So this is code I copied from semver-parser (until https://github.com/steveklabnik/semver-parser/issues/45 is addressed).  I've tried to keep it closer to the original, so I left it with `try!` rather than converting it over.

Looks like they made `try` a reserved word in 2018 edition, so this is using the [raw identifer syntax](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/raw-identifiers.html) to still be able to access the `try!` macro.

Issue #93, opened 2019-03-11
Should we support cascading configurations?


epage:
The idea being that we merge the configs from the following locations
- CLI flags
- manifest path's directory
- either every parent directory or the workspace root
- home
- [env variables](https://crates.io/crates/envy) (where does this fall in precedent order?)
- defaults

Use cases
- Centralize most release knowledge with overrides per repo in a mono repo
- Allow some settings to be left up to the user

Questions
- Should replacement file paths be relative to the manifest path or the release config?

Possible implementation
- Switch all config settings to `Option<_>`
- Have a `merge` function
- Have member functions that `unwrap_or` the field's default

Alternative implementation
- Load it all into a hash map, perform hash map merges, and then move to the current config struct.  The downside is this defers error reporting, making it so you can't report which file caused the problem

sunng87:
Currently we support the config source from:

* cli options: top priority
* `release.toml` in project root
* `package.metadata.release` section in `Cargo.toml`
* `.release.toml` under `HOME`: used when neither `release.toml` nor `package.metadata.release` configured

As we have workspace supported, it makes sense to support `release.toml` or `meradata.release` from workspace parent or root. 

> Should replacement file paths be relative to the manifest path or the release config? 

I think it should be relative to where it configured. 

> Have a merge function

Good catch because items in replacement should be concat together instead of overwrited.

I'm +1 for the first implementation proposal. It may also simplify [the big section](https://github.com/sunng87/cargo-release/blob/master/src/main.rs#L70-L145) in `main.rs`

epage:
> /// 2. $(pwd)/Cargo.toml `package.metadata.release` (with deprecation warning)

I'm not seeing this deprecation warning.

PR #110, opened 2019-05-06
Merge args into config

epage:
This is preparation for #93.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -7,55 +7,289 @@ use serde::{Deserialize, Serialize};
 
 use crate::error::FatalError;
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+pub trait ConfigSource {
+    fn sign_commit(&self) -> Option<bool> {
+        None
+    }
+
+    fn upload_doc(&self) -> Option<bool> {
+        None
+    }
+
+    fn push_remote(&self) -> Option<&str> {
+        None
+    }
+
+    fn doc_branch(&self) -> Option<&str> {
+        None
+    }
+
+    fn disable_publish(&self) -> Option<bool> {
+        None
+    }
+
+    fn disable_push(&self) -> Option<bool> {
+        None
+    }
+
+    fn dev_version_ext(&self) -> Option<&str> {
+        None
+    }
+
+    fn no_dev_version(&self) -> Option<bool> {
+        None
+    }
+
+    fn pre_release_commit_message(&self) -> Option<&str> {
+        None
+    }
+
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -7,55 +7,349 @@ use serde::{Deserialize, Serialize};
...
-            tag_prefix: None,
-            doc_commit_message: "(cargo-release) generate docs".into(),
-            disable_tag: false,
-            enable_features: vec![],
-            enable_all_features: false,
-            dependent_version: Default::default(),
+impl Config {
+    pub fn update(&mut self, source: &ConfigSource) {
+        if let Some(sign_commit) = source.sign_commit() {
+            self.sign_commit = Some(sign_commit);
```
sunng87: why not just `self.sign_commit = source.sign_commit().clone()`

epage: Either way. Another option is to encapsulate this pattern with an extension trait.

epage: Thinking more about this...

The current approach references the struct twice with the proposed referencing it three times which increases the chance of error.

sunng87: Or is it possible to write a procedural macro for this? To avoid multiples changes required for adding new config option.

sunng87:
@epage do you want to improve on this branch, or I will merge it to let the process go

src/config.rs
epage: Yes, a procedural macro could help here.  Maybe I'm just not familiar enough with them but it seems like a big hammer to solve this small problem. 

I do want to take these lessons back to the CLI-WG [discussions](https://github.com/rust-lang-nursery/cli-wg/issues/7).  It feels like the current selection of crates is too all-in-one which doesn't let us adapt for the needs of crates like this one.  I'm thinking a generalized set of macros to let people layer their config as needed would be helpful for the community.

epage:
> @epage do you want to improve on this branch, or I will merge it to let the process go

If you are fine with it, I'd say let's get this in and continue to iterate.

sunng87:
Yes.

src/config.rs
sunng87: I had a [clojure library](https://github.com/sunng87/stavka), which I think deals this well. I'll think about the challenge to port it to static-typed language like rust.

epage: `config` and `configure` do it to a degree but I think they both use a `HashMap` when merging the sources which delays error reporting to the end.  My goal with this was to report the error (invalid data type, extra field, etc) with being able to tell the user where the problem came from so they could fix it.

PR #111, opened 2019-05-08
Layered config support

### LINKED ISSUE #93: Should we support cascading configurations?

epage:
Includes loading a workspace-wide config.  To make hooks still work, this also adds environment variables to access the workspace and crate paths.

Fixes #93

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,6 +10,7 @@ use structopt;
 #[cfg(test)]
 extern crate assert_fs;
 
+use std::ffi::OsStr;
 use std::path::Path;
 use std::process::exit;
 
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -270,9 +270,11 @@ fn execute(args: &ReleaseOpt) -> Result<i32, error::FatalError> {
...
             shell::log_info(&format!("Calling pre-release hook: {:?}", pre_rel_hook));
             let envs = btreemap! {
-                "PREV_VERSION" => prev_version_string.as_ref(),
-                "NEW_VERSION" => new_version_string.as_ref(),
-                "DRY_RUN" => if dry_run { "true" } else { "false" }
+                OsStr::new("PREV_VERSION") => prev_version_string.as_ref(),
+                OsStr::new("NEW_VERSION") => new_version_string.as_ref(),
+                OsStr::new("DRY_RUN") => OsStr::new(if dry_run { "true" } else { "false" }),
+                OsStr::new("WORKSPACE_ROOT") => ws_meta.workspace_root.as_os_str(),
+                OsStr::new("CRATE_ROOT") => manifest_path.parent().unwrap_or_else(|| Path::new(".")).as_os_str(),
```
epage: I need to add some tests to make sure this actually is useful for the various use cases.

I suspect it won't be and that we'll need to do replacements on the hook to allow it to be shared.

I decided against tracking the directory that defines the hook and setting the CWD to simplify the merging process.

src/main.rs
```diff
@@ -407,7 +409,7 @@ struct ReleaseOpt {
     level: version::BumpLevel,
 
     #[structopt(short = "c", long = "config")]
-    /// Custom config file
+    /// Custom config file (isolated)
     config: Option<String>,
```
epage: I think I actually want to merge the custom config in with the rest and add an `--isolated` that will cause us to ignore everything but the custom config.

PR #115, opened 2019-06-01
 docs: Split up docs 

epage:
The goal is to make it easier for people to find the documentation
relevant to their needs.


epage:
See https://github.com/epage/cargo-release/tree/docs to explore what this is like.

sunng87:
Perhaps we can also include the content of `reference.md` into `cargo release --help`

epage:
I've been curious about the idea for a crate that would render markdown on the terminal.

sunng87:
Markdown has excellent readability, it's ok to dump it directly I think

Issue #119, opened 2019-06-04
Crafting a release


sunng87:
Another scenario I have been thinking about is to support release proposal. I'm not support if cargo-release really wants to cover the whole process. Typically it's like:

* going to propose a release
* prepare release notes
* `cargo prepare-release`: 
  * create a release branch and bump version
  * create a github pull request titled like "Vote: Release 1.0.0"
* pull request approved and merged
* travis or other CIs received notification and run `cargo perform-release` to actually merge the release branch and publish to crates.io

This process could work for projects maintained by several devs or a committee. At the moment, `cargo-release` only works well for personal development.

lucab:
You may be interested in how we (CoreOS/RedHat) perform team-reviewed releases using `cargo-release`:
 * https://github.com/coreos/afterburn/blob/master/docs/release-guide.md
 * https://github.com/coreos/afterburn/issues?q=Release+author%3Alucab

sunng87:
Thanks @lucab !

Your process is almost compatible with the one I described. I might not support github pull request/issue creation (possible to hook into `prepare-release`), but it's definitely possible to cover git branching in cargo-release. Thanks for letting me know your case.

passcod:
I've got this kind of workflow for my own crates, and I've just made a github action that provides support for it: https://github.com/cargo-bins/release-pr

epage:
With the upcoming 0.22 release, users will be able to create their own two-phase release

Workflow 1

Assuming the CHANGELOG is always valid

A contributor runs:
```console
$ cargo release version patch -x
$ git commit -a -m "..."
```
and creates a PR.   CI runs `cargo release --unpublished`.  If it fails with code=2, then nothing needs to release.  If it succeeds, there is something to release.

The PR gets merged.

Pushes to `main` trigger running `cargo release --unpublished -x`.  This runs replacements, commits, publishes, tags, and pushes.

Workflow 2

A contributor runs:
```console
$ cargo release patch -x
$ cargo release replace -x
$ git commit -a -m "..."
```
and creates a PR.  CI runs `cargo release publish --workspace && cargo release tag --workspace && cargo release push --workspace` to show what will be published, tagged, and pushed.  We'll automatically skip things that don't need to be published or tagged.

The PR gets merged.

Pushes to `main` trigger running `cargo release publish --workspace && cargo release tag --workspace && cargo release push --workspace`.  That last push should only push tags, not branches.

Neither of these workflows work with `dev-version-ext`.

[...]

PR #127, opened 2019-06-16
(feat) add v to default version tag

### LINKED ISSUE #124: Add `v` into the default tag format?
Currently, the tag format is `{{prefix}}{{version}}` but I tend to more so see `{{prefix}}v{{version}}` within the rust community.  This makes sense because it makes it easier to do [a search to find version tags](https://github.com/japaric/trust/blob/master/ci/install.sh#L34).

#123 is about making the format customizable.  This is about changing the default, regardless of it being customizable.

sunng87:
Fixes #124

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -260,7 +260,7 @@ impl Config {
         self.tag_name
             .as_ref()
             .map(|s| s.as_str())
-            .unwrap_or("{{prefix}}{{version}}")
+            .unwrap_or("{{prefix}}v{{version}}")
     }
 
     pub fn doc_commit_message(&self) -> &str {


epage:
We should remember to note in a changelog that we've broken compatibility but that people can revert to the old behavior be setting the tag-name.

sunng87:
hi @epage I just invited you as owner of cargo-release on crate.io, did you receive the invitation?

It seems with two owners on a crate, we can move forward on scenario described in #119

epage:
Accepted.  Didn't get a notification but did see it when I knew to look for it.

Issue #117, opened 2019-06-03
What would an auto-release workflow look like?


epage:
Something I've been considering is the idea of auto-releasing after each PR but the question is "how?"
- Determine bump level
- Ensure we don't get into an infinite loop (a release causes a commit which causes a release)
- Any changelist management?

epage:
One approach I just came across via [awesome-rust](https://github.com/rust-unofficial/awesome-rust#development-tools) is [semantic](https://github.com/semantic-rs/semantic-rs/blob/master/src/main.rs) which uses clog to generate the changelog and determine the bump level based on the commit messages.

sunng87:
Personally I prefer an isolation between commit log, which is dev-facing, and release notes, which is user facing. So I didn't included auto changelog generation from very beginning. 

However, I agree it's possible to determine the release level from commit log or public api changes. I wish semantic could offer a binary of library just for determining bump level so we can add optional support for that.

epage:
> Personally I prefer an isolation between commit log, which is dev-facing, and release notes, which is user facing. So I didn't included auto changelog generation from very beginning.

Ideally, yes but I maintain too many crates to always worry about the changelog.  Maybe switching to cargo-release will help.  However, I do tend to make my commits more user focused (if they are feat/fix rather than refactor/chore).  

And too often I see projects, major projects, without them and I'm at a loss digging through commits when they make breaking changes.  Something is better than nothing.

> However, I agree it's possible to determine the release level from commit log or public api changes. I wish semantic could offer a binary of library just for determining bump level so we can add optional support for that.

At my quick glance, it looked like semantic was just calling into `clog-cli`'s crate to see if anything was considered breaking from conventional-changelog's perspective, so that should be relatively easy to pull out.

epage:
The downside to `semantic`'s approach is you have to hope you did things right for it to correctly guess what version field to bump.  If you mess up, you might end up needing to yank a release.  I have also found that while it is easy to get people to do checklist tasks in the source code (like update changelog), it is harder to be programmatically precise tasks like with commit messages.

Another approach is to instead have the source always represent what the next version will be.  If a commit makes a breaking change, then that commit is responsible for updating the version number without doing a release.  The CI then releases that version.  To support this, we might want a `--no-release` flag that is a shortcut for `--no-push`, `--no-tag`, etc that a developer will call in prep for their PR.

The use cases we'd want to consider supporting
- Dev bumps minor, major with or without pre-release
  - `--no-release` would help
[...]

PR #131, opened 2019-07-25
Add default-level config

dalance:
This PR add `default-level` to config file.

In https://github.com/sunng87/cargo-release/issues/77, workspace with pre-release does not seem to be supported.
So I want to use `cargo release patch` with `no-dev-version` option.
If there is `default-level` configuration, I can `cargo release` simply.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -57,6 +57,7 @@ Configuration is read from the following (in precedence order)
 | `pre-release-hook` | \-          | list of arguments | Provide a command to run before `cargo-release` commits version change. If the return code of hook command is greater than 0, the release process will be aborted. |
 | `enable-features` | `--features` | list of names | Provide a set of feature flags that should be passed to `cargo publish` (requires rust 1.33+) |
 | `all-features` | `--all-features` | bool  | Signal to `cargo publish`, that all features should be used (requires rust 1.33+) |
+| `default-level` | \- | string | Default bump level |
 
 ### Pre-release Replacements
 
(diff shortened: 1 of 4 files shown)


sunng87:
Thanks @dalance . I think in real world your bump version is determined by the changes in the release. Adding a config item to make it constant default may not fit actual usage and can be misleading further. @epage WDTY

dalance:
Thank you for your comment.

I think `cargo release` is most frequently used command.
But if `no-dev-version` is used, `cargo release` becomes meaningless. I feel this behavior is inconsistent.
In this viewpoint, if default bump level with `no-dev-version` becomes patch, it may be consistent because default increment of pre-release is patch version.

epage:
Regarding the discussion in #77, I created https://github.com/sunng87/cargo-release/issues/132 in response.

epage:
As for this request, its a question of workflows and common cases

Currently, the default level works well for the workflows where
- pre-release is used
- crate is already at the desired version

This involves another potential workflow where pre-release isn't used so the min change needed is patch.

The workflow seems reasonable but questions:
- Is creating a patch release common enough to warrant this?
- Are there other reasons for us to make a default, like related to #117 (I think instead that calls for a `level` for post-release)
- Since the crux of the issue is #132, should the focus instead be on that?

dalance:
pre-release flow seems to be recommended.
If so, I think pre-release support of workspace by resolving #132 has higher priority than this workaround.

PR #136, opened 2019-08-02
Use {{next_version}} in default for pro_release_commit_message

dead10ck:
The default value for `pro_release_commit_message` was using `{{version}}` instead of `{{next_version}}`.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -231,7 +231,7 @@ impl Config {
         self.pro_release_commit_message
             .as_ref()
             .map(|s| s.as_str())
-            .unwrap_or("(cargo-release) start next development iteration {{version}}")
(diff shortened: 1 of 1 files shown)


epage:
Looking into this to confirm what the behavior should be, it looks like [this commit](https://github.com/sunng87/cargo-release/commit/dcfa3b1a0ec597c8c9c978680666e77fa0f100d6#diff-639fbc4ef05b315af92b4d836c31b023) introduced the split between `{{version}}` and `{{next_version}}` but the default never got updated.

I was surprised we didn't catch this with our own history, so I assumed something must be off [but instead our history does read in a weird way](https://github.com/sunng87/cargo-release/commit/8598453570d3f8c0244392ebcabe8780a7b23862)

sunng87:
This seems to be a regression issue introduced in our refactoring.

PR #139, opened 2019-08-13
Support git tag "--edit" option

zazabe:
In our use-case, we want to add the changelog manually as the tag message, we can easily do it by adding `git tag` [--edit](https://git-scm.com/docs/git-tag#Documentation/git-tag.txt---edit) support.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -60,6 +60,8 @@ pub trait ConfigSource {
         None
     }
 
+    fn tag_edit(&self) -> Option<bool> { None }
+
     fn tag_prefix(&self) -> Option<&str> {
         None
     }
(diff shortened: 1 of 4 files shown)


sunng87:
Thanks @zazabe for this patch! The first idea about this was if we offer a tag edit option, why not also add commit message editing? It will be more consistency in terms of customization.

src/config.rs
```diff
@@ -162,6 +167,9 @@ impl Config {
         if let Some(tag_name) = source.tag_name() {
             self.tag_name = Some(tag_name.to_owned());
         }
+        if let Some(tag_edit) = source.tag_edit() {
+            self.tag_edit = Some(tag_edit);
+        }
```
matthiasbeyer: Couldn't this be written as simple as `self.tag_edit = source.tag_edit()`?

epage: The idea behind this function is to merge configuration files.  If `source` doesn't have a setting, than we should use `self`s.

If we did `self.tag_edit = source.tag_edit()`, then a `None` would overwrite a `Some`.

matthiasbeyer: Ah, sure! My bad!

epage:
I'm not convinced interactive editing of tags and commits should be in the config.  Could you open an issue for us to better explore the requirements for this and how it fits within cargo-release?

PR #140, opened 2019-08-17
Rename "pro-release-commit-message" to "post-release-commit-message"

### LINKED ISSUE #137: "pro-release-commit-message" is a confusing name
Currently there are two config parameters: `pre-release-commit-message` and `pro-release-commit-message`. At first glance, the second one looks like a typo, but it is not. I think the intended name might have been `post-release-commit-message`, which would more accurately convey what the config option is.

sunng87:
Fixes #137 

This patch rename `pro-release-commit-message` to `post-release-commit-message` while maintain backward compatibility for now.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -44,7 +44,12 @@ pub trait ConfigSource {
         None
     }
 
+    // depreacted
     fn pro_release_commit_message(&self) -> Option<&str> {
+        None
+    }
+
+    fn post_release_commit_message(&self) -> Option<&str> {
         None
     }
 
@@ -102,7 +107,9 @@ pub struct Config {
     pub dev_version_ext: Option<String>,
     pub no_dev_version: Option<bool>,
     pub pre_release_commit_message: Option<String>,
+    // depreacted
     pub pro_release_commit_message: Option<String>,
+    pub post_release_commit_message: Option<String>,
     pub pre_release_replacements: Option<Vec<Replace>>,
     pub pre_release_hook: Option<Command>,
     pub tag_message: Option<String>,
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -144,8 +151,13 @@ impl Config {
         if let Some(pre_release_commit_message) = source.pre_release_commit_message() {
             self.pre_release_commit_message = Some(pre_release_commit_message.to_owned());
         }
+        // depreacted
         if let Some(pro_release_commit_message) = source.pro_release_commit_message() {
-            self.pro_release_commit_message = Some(pro_release_commit_message.to_owned());
+            println!("[warning] pro_release_commit_message is deprecated, use post-release-commit-message instead");
```
epage: We should eventually switch to `log` crate.

src/config.rs
```diff
@@ -144,8 +151,13 @@ impl Config {
...
+            self.post_release_commit_message = Some(pro_release_commit_message.to_owned());
```
epage: Having the compatibility code in the merge layer seems brittle.  Slight changes to the merging logic and we would lose compatibility.

My first thought is to modify `impl ConfigSource for Config` to `or_else` the pro-release into the post-release message and remove pro-release form the trait.

That also suffers from the merge issue and we'd need to also modify `impl Config` to do the same thing.

Now I'm wondering why I didn't implement `impl Config` in terms of `impl ConfigSource` since they have to duplicate all the line noise for switching to reference types anyways.

sunng87: The idea to put compatibility code in config module is to keep upper logic clean. 

I found there is no write method in `Config` because it's directly mapped from data so for now I think the merging layer was best place to control how to use these properties.

sunng87:
Thanks @epage I will merge this for now.

src/config.rs
epage: To be clear, I was still saying to put it in the config module, just at the layer where the information is provided for merging rather than the merge code itself. But I don't think this is a reason to block and have no problem that this was merged.

PR #148, opened 2019-09-19
 feat(git): Notify user on unchanged crates 

### LINKED ISSUE #143: Update crates only if they changed
Does (or could) `cargo release` check whether files have been changed since the last release before tagging a new version?

I have a workspace with a bunch of different crates that depend on one another. It would be nice if this tool would tag and release new versions of the sub-crates only when they've actually changed since the last time they were released.

Also, thank you for your work on this project!

epage:
Ideally, we could give people the option to skip unchanged crates.
We're starting conservativly though in case there are changes that won't
always be detected this way (like updates to dependencies). The big
concern is making this smart enough that people stop thinking while
still being dumb enough that people should still think.

For now, this incorrectly identifies crates as unchanged during dry-run
if we'd update its dependencies as part of releasing a dependency that
lives in the workspace.

For now, this does not respect `package.exclude`, so changes may occur
that should not force the crate to be released.

Fixes #143

BREAKING CHANGE: `{{tag_prefix}}` is now calculated before `{{version}}`
is updated to the current version and should not be trusted.

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -10,6 +10,7 @@ pub fn is_dirty(dir: &Path) -> Result<bool, FatalError> {
         .arg("diff")
         .arg("HEAD")
         .arg("--exit-code")
+        .arg("--name-only")
         .current_dir(dir)
         .output()
         .map_err(FatalError::from)?;
--- a/src/git.rs
+++ b/src/git.rs
@@ -29,6 +29,23 @@ pub fn is_dirty(dir: &Path) -> Result<bool, FatalError> {
     Ok(tracked_unclean || untracked)
 }
 
+pub fn changed_from(dir: &Path, tag: &str) -> Result<Option<bool>, FatalError> {
+    let output = Command::new("git")
+        .arg("diff")
+        .arg(&format!("{}..HEAD", tag))
+        .arg("--name-only")
+        .arg("--exit-code")
+        .arg(".")
+        .current_dir(dir)
+        .output()
+        .map_err(FatalError::from)?;
+    match output.status.code() {
+        Some(0) => Ok(Some(false)),
+        Some(1) => Ok(Some(true)),
+        _ => Ok(None), // For cases like non-existent tag
+    }
+}
+
 pub fn commit_all(dir: &Path, msg: &str, sign: bool, dry_run: bool) -> Result<bool, FatalError> {
     call_on_path(
         vec!["git", "commit", if sign { "-S" } else { "" }, "-am", msg],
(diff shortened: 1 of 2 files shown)


epage:
> For now, this incorrectly identifies crates as unchanged during dry-run
if we'd update its dependencies as part of releasing a dependency that
lives in the workspace.

I plan to file a bug about this when this is merged.

> For now, this does not respect package.exclude, so changes may occur
that should not force the crate to be released.

I plan to open an enhancement issue for this when merged.

epage:
@emschwartz any suggestions on this to help with your workflow?

emschwartz:
Thanks for working on this!

Would it be possible to prompt the user about whether they want to bump the
version when it thinks nothing has changed? Or, is there another way for me
to use the logged message to make sure not to bump the version of that
crate?

epage:
> Would it be possible to prompt the user about whether they want to bump the
version when it thinks nothing has changed? Or, is there another way for me
to use the logged message to make sure not to bump the version of that
crate?

- If you first do a dry-run, you can then do an `--exclude`.  Note that the information during a dry-run is inaccurate if, in a workspace, you have dependencies to upgrade within the workspace
- If you run this without a dry-run, the warning is before the confirmation prompt.  I believe the confirmation prompt will cause it to immediately exit.  You'd need to re-run with the no-change and all dependencies' `--exclude`ed

Potential improvements:
- Consider not exiting if you say "no" to the current crate
- Fix the dry-run (will require re-arranging things a bit)

sunng87:
While the patch looks good me, I just came across an idea that if a crate can be unchanged during a workspace release, what if crates in workspace has different release levels? For example, if `serde-json` and `serde` shares a workspace, there can be breaking change for `serde-json`'s JSON API but `serde`'s API remains unchanged. 

Previously I suppose crates in a workspace shares a common version.

epage:
> While the patch looks good me, I just came across an idea that if a crate can be unchanged during a workspace release, what if crates in workspace has different release levels? For example, if serde-json and serde shares a workspace, there can be breaking change for serde-json's JSON API but serde's API remains unchanged.

Yup, been keeping this use case in mind when I've been working on workspace features.  For example, [predicates](https://github.com/assert-rs/predicates-rs) separates out the less stable stuff from the stable (which lives in a `-core` crate).  This was important since other crates would have `predicates-core` in their public API but not `predicates`, making them break compatibility less often.

This can be helped by the fact that we can call `cargo-release` on different subsets of the workspace and the fact that `cargo-release` offers relative bumps (a "minor" change to a workspace can equally bump a 1.0 and a 2.0). 

Something I've been considering is adding support for absolute version bumps (3.0.0 instead of "major").  My use case is for when I do want to bump an entire workspace when they've previously been bumped inconsistently, I'll want to re-sync up on the versions (maybe?)

PR #151, opened 2019-09-20
Polish CLI arguments

epage:
- Support cargo's replacement for `--all`, `--workspace`
- Support `-n` for dry-run

And in general, this sees dependencies being updated.


epage:
I'm assuming this is relatively uninteresting and will merge it.  Let me know if there are concerns; we can always iterate in future PRs.

PR #152, opened 2019-09-21
fix(output): Switch to logging module

epage:
This will give the user more conrol over what is outputed.

The default log level is now "error".  Before it was:
- For dry-run: "trace"
- For all else: "debug"


epage:
I'm assuming this is relatively uninteresting and will merge it. Let me know if there are concerns; we can always iterate in future PRs.

PR #159, opened 2019-10-08
Skip git push if remote is not configured [#155]

### LINKED ISSUE #155: Not having an upstream repo set should not fail
The command that cargo release does is `git push`, which assumes that an upstream repo is set for the current branch. It will fail if this is not the case. Having an upstream branch set should not be required.

I think the logic could go something like this:

1. If the upstream remote for the current branch *is* set, then just run `git push`
1. If the upstream remote for the current branch *is not* set *and* the `push-remote` config *is* set, then run `git push <push-remote> <current branch>`
1. If neither the upstream remote for the current branch nor the `push-remote` config is set set, then run `git push origin <current branch>`

sunng87:
Added a check before running `git push`. If the remote is not configured, print info message and skip the push.

Fixes #155

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -469,11 +469,18 @@ fn release_package(
     if !pkg.config.disable_push() {
         log::info!("Pushing to git remote");
         let git_remote = pkg.config.push_remote();
-        if !git::push(cwd, git_remote, dry_run)? {
-            return Ok(106);
-        }
-        if !pkg.config.disable_tag() && !git::push_tag(cwd, git_remote, &tag_name, dry_run)? {
-            return Ok(106);
+        if git::check_remote(cwd, git_remote) {
+            if !git::push(cwd, git_remote, dry_run)? {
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -469,11 +469,18 @@ fn release_package(
...
+        if git::check_remote(cwd, git_remote) {
+            if !git::push(cwd, git_remote, dry_run)? {
+                return Ok(106);
+            }
+            if !pkg.config.disable_tag() && !git::push_tag(cwd, git_remote, &tag_name, dry_run)? {
+                return Ok(106);
+            }
+        } else {
+            log::info!(
+                "push_remote: {} is not configured for git, skip git push",
```
epage: Is this an info, warning, or error?

I feel like for each user, it'll be different :(  Users can disable pushing though, so maybe I lean towards at least warn?

epage:
Superseded by #292

PR #163, opened 2019-10-11
 refactor: Pre-compute some information 

epage:
My goal is to reduce the release logic to eventually allow
- Doing "anything unchanged" checks before the confirmation prompt
- Include dependency updates  in "anything unchanged" when running in
  dry-run mode
- A single confirmation prompt
- Potentially consolidate commits
- Consolidate pushes

This stopped short of the refactorings needed to unlock these changes
because it was getting to big.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -22,7 +22,7 @@ use semver::Identifier;
 use structopt::StructOpt;
 
 use crate::error::FatalError;
-use crate::replace::{do_file_replacements, replace_in, Replacements};
+use crate::replace::{do_file_replacements, Template};
 
 mod cargo;
 mod cmd;
@@ -54,18 +54,36 @@ struct Package<'m> {
     manifest_path: &'m Path,
     package_path: &'m Path,
     config: config::Config,
+
+    is_root: bool,
+
+    prev_version: Version,
+    prev_tag: String,
+    version: Option<Version>,
+    post_version: Option<Version>,
+
+    //dependent_version: config::DependentVersion,
+    //dependents: Vec<&'m Path>,
+    //failed_dependents: Vec<&'m Path>,
+    features: Features,
+}
+
+struct Version {
+    version: semver::Version,
+    version_string: String,
 }
 
 impl<'m> Package<'m> {
     fn load(
         args: &ReleaseOpt,
+        git_root: &Path,
         ws_meta: &'m cargo_metadata::Metadata,
         pkg_meta: &'m cargo_metadata::Package,
     ) -> Result<Self, error::FatalError> {
         let manifest_path = pkg_meta.manifest_path.as_path();
         let cwd = manifest_path.parent().unwrap_or_else(|| Path::new("."));
 
-        let release_config = {
+        let config = {
             let mut release_config = config::Config::default();
 
             if !args.isolated {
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -97,11 +115,86 @@ impl<'m> Package<'m> {
...
+        let prev_tag = {
+            let mut template = Template {
+                prev_version: Some(&prev_version.version_string),
+                version: Some(&prev_version.version_string),
+                crate_name: Some(pkg_meta.name.as_str()),
+                ..Default::default()
+            };
+
+            let tag_prefix = config.tag_prefix(is_root);
+            let tag_prefix = template.render(tag_prefix);
```
epage: We no longer include `date` when rendering `tag_prefix`

src/main.rs
```diff
@@ -97,11 +115,86 @@ impl<'m> Package<'m> {
...
+            template.prefix = Some(&tag_prefix);
+            template.render(config.tag_name())
```
epage: We no longer include `date` when rendering `tag_name`

src/main.rs
```diff
@@ -428,37 +492,57 @@ fn release_package(
...
+    } else {
+        let base = pkg.version.as_ref().unwrap_or_else(|| &pkg.prev_version);
+        let mut template = Template {
+            prev_version: Some(&pkg.prev_version.version_string),
+            version: Some(&base.version_string),
+            crate_name: Some(crate_name),
+            ..Default::default()
+        };
+        let tag_prefix = pkg.config.tag_prefix(pkg.is_root);
+        let tag_prefix = template.render(tag_prefix);
```
epage: We no longer include `date` when rendering `tag_prefix`

src/main.rs
```diff
@@ -428,37 +492,57 @@ fn release_package(
...
+        template.prefix = Some(&tag_prefix);
+
+        let tag_name = template.render(pkg.config.tag_name());
```
epage: We no longer include `date` when rendering `tag_name`

src/main.rs
```diff
@@ -428,37 +492,57 @@ fn release_package(
...
+        let base = pkg.version.as_ref().unwrap_or_else(|| &pkg.prev_version);
+        let template = Template {
+            prev_version: Some(&pkg.prev_version.version_string),
+            version: Some(&base.version_string),
+            crate_name: Some(crate_name),
+            date: Some(now.as_str()),
+            next_version: Some(updated_version_string),
+            ..Default::default()
+        };
+        let commit_msg = template.render(pkg.config.post_release_commit_message());
```
epage: We no longer include `prefix` / `tag_name` when rendering `commit_msg`

src/main.rs
```diff
@@ -378,7 +450,14 @@ fn release_package(
...
 
-        let commit_msg = replace_in(pkg.config.pre_release_commit_message(), &replacements);
+        let template = Template {
+            prev_version: Some(&pkg.prev_version.version_string),
+            version: Some(&new_version_string),
+            crate_name: Some(crate_name),
+            date: Some(now.as_str()),
+            ..Default::default()
+        };
+        let commit_msg = template.render(pkg.config.pre_release_commit_message());
```
epage: We no longer include `prefix` / `tag_name` when rendering `commit_msg`

src/main.rs
```diff
@@ -347,9 +412,16 @@ fn release_package(
 
         if !pkg.config.pre_release_replacements().is_empty() {
             // try replacing text in configured files
+            let template = Template {
+                prev_version: Some(&pkg.prev_version.version_string),
+                version: Some(&new_version_string),
+                crate_name: Some(crate_name),
+                date: Some(now.as_str()),
+                ..Default::default()
```
epage: This didn't have `prefix` / `tag_name` before when doing file replacements.

src/main.rs
```diff
@@ -97,11 +115,86 @@ impl<'m> Package<'m> {
...
+            let tag_prefix = template.render(tag_prefix);
+            template.prefix = Some(&tag_prefix);
+            template.render(config.tag_name())
+        };
+
+        let version = {
+            let mut potential_version = prev_version.version.clone();
+            if args
+                .level
+                .bump_level(&mut potential_version, args.metadata.as_ref())?
```
epage: This call is erroring.  I need to look into why.

src/main.rs
epage: This is aligned with the documentation
https://github.com/sunng87/cargo-release/blob/master/docs/reference.md#placeholders

src/main.rs
epage: Thoughts on if `{{date}}` should be exposed for these renders?

sunng87: In some case `date` is used as part of version string, but for tag I can't see the need for `{{date}}` alone.

PR #169, opened 2019-10-12
 fix(log): Change default level to INFO

### LINKED ISSUE #161: Changing default logging level to INFO?
After introducing `log` for logging, the default level doesn't print any information during `cargo-release`, made it difficult to know what's going on and diagnose when error encountered. 

If we change level to `INFO`, which meanings `-vv` is the default, we probably need an inverted option, for example `-q` to reduce log to `WARN` and `-qq` to `ERROR`. I didn't know if any app using `-q` like this. Maybe there is better solution for this?

epage:
Fixes #161

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -625,23 +625,29 @@ struct ReleaseOpt {
 
 #[derive(StructOpt, Debug, Clone)]
 pub struct Verbosity {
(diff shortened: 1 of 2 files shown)


sunng87:
That's fast!

PR #175, opened 2019-10-31
Batch process each step

epage:
Instead of processing one crate at a time across all steps, we're going to process one step at a time across all crates.   I have no idea if, in the end, this will make it easier or harder to recover.  The main goal is to work towards:
- Doing "anything unchanged" checks before the confirmation prompt
- Include dependency updates  in "anything unchanged" when running in
  dry-run mode, https://github.com/sunng87/cargo-release/issues/150
- A single confirmation prompt, https://github.com/sunng87/cargo-release/issues/138
- Potentially consolidate commits, https://github.com/sunng87/cargo-release/issues/138
- Consolidate pushes, https://github.com/sunng87/cargo-release/issues/138

Besides how things are processed, this does change
- A single push of HEAD is done rather than once per crate (we still push once per tag)
- We confirm once
- We confirm regardless of version change happening.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -33,6 +33,9 @@ mod replace;
 mod shell;
 mod version;
 
+static NOW: once_cell::sync::Lazy<String> =
+    once_cell::sync::Lazy::new(|| Local::now().format("%Y-%m-%d").to_string());
+
 fn find_dependents<'w>(
     ws_meta: &'w cargo_metadata::Metadata,
     pkg_meta: &'w cargo_metadata::Package,
@@ -339,7 +342,6 @@ fn release_package(
     let sign = pkg.config.sign_commit();
     let cwd = pkg.package_path;
     let crate_name = pkg.meta.name.as_str();
-    let now = Local::now().format("%Y-%m-%d").to_string();
 
     // STEP 2: update current version, save and commit
     if let Some(version) = pkg.version.as_ref() {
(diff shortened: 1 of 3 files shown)


sunng87:
@epage is this ready for review and merge?

epage:
Yes, sorry, I forgot to request your review.

src/main.rs
```diff
@@ -394,184 +442,191 @@ fn release_package(
...
-        if !git::push(cwd, git_remote, dry_run)? {
-            return Ok(106);
-        }
-        if let Some(tag_name) = pkg.tag.as_ref() {
-            if !git::push_tag(cwd, git_remote, &tag_name, dry_run)? {
-                return Ok(106);
+    let mut pushed: HashSet<_> = HashSet::new();
+    for pkg in pkgs {
+        if !pkg.config.disable_push() {
+            let git_remote = pkg.config.push_remote();
```
sunng87: perhaps we can `git push` only once?

for tags, it has to be pushed individually.

src/main.rs
```diff
@@ -394,184 +442,191 @@ fn release_package(
...
-
-        log::info!("Commit and push docs for {}", crate_name);
-        git::init(&doc_path, dry_run)?;
-        git::add_all(&doc_path, dry_run)?;
-        git::commit_all(&doc_path, pkg.config.doc_commit_message(), sign, dry_run)?;
-        let default_remote = git::origin_url(cwd)?;
-
-        let refspec = format!("master:{}", pkg.config.doc_branch());
-        git::force_push(&doc_path, default_remote.trim(), &refspec, dry_run)?;
+    for pkg in pkgs {
```
sunng87: we can remove this deprecated doc generation step

src/main.rs
epage: Technically, each package could have a different remote configured, so we `push` once per remote (see `if !pushed.contains(git_remote)` on the next line.

src/main.rs
epage: Probably should be a separate PR (I can do it right after this one).

sunng87: no problem!

PR #180, opened 2019-11-07
Add allow-dirty flag

ArminBecher:
Add flag to allow releases with a dirty repository. Analog to the other cargo calls such as `cargo publish`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 Performs release best-practices, including:
 
-* Ensure the git working directory is clean.
+* Ensure the git working directory is clean (except if the `--allow-dirty` flag is set).
 * Bump the version in Cargo.toml
 * Run `cargo publish` ([if not disabled](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish--field-optional))
 * Create a git tag for this version
--- a/src/main.rs
+++ b/src/main.rs
@@ -242,7 +242,7 @@ fn release_workspace(args: &ReleaseOpt) -> Result<i32, error::FatalError> {
     let ws_meta = args.manifest.metadata().exec().map_err(FatalError::from)?;
 
     git::git_version()?;
-    if git::is_dirty(&ws_meta.workspace_root)? {
+    if !args.allow_dirty && git::is_dirty(&ws_meta.workspace_root)? {
         log::warn!("Uncommitted changes detected, please commit before release.");
         if !args.dry_run {
             return Ok(101);
@@ -548,7 +548,7 @@ fn release_packages<'m>(
             log::info!("Running cargo publish on {}", crate_name);
             // feature list to release
             let features = &pkg.features;
-            if !cargo::publish(dry_run, &pkg.manifest_path, features)? {
+            if !cargo::publish(dry_run, args.allow_dirty, &pkg.manifest_path, features)? {
                 return Ok(103);
             }
         }
@@ -673,7 +673,7 @@ struct ReleaseOpt {
     custom_config: Option<String>,
 
     #[structopt(long)]
-    /// Ignore implicit configuration files.
+    /// Ignore implicit configuration files
     isolated: bool,
 
     #[structopt(flatten)]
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -6,7 +6,7 @@
 
 Performs release best-practices, including:
 
-* Ensure the git working directory is clean.
+* Ensure the git working directory is clean (except if the `--allow-dirty` flag is set).
```
epage: Thanks for considering the impact on documentation!

src/main.rs
```diff
@@ -242,7 +242,7 @@ fn release_workspace(args: &ReleaseOpt) -> Result<i32, error::FatalError> {
     let ws_meta = args.manifest.metadata().exec().map_err(FatalError::from)?;
 
     git::git_version()?;
-    if git::is_dirty(&ws_meta.workspace_root)? {
+    if !args.allow_dirty && git::is_dirty(&ws_meta.workspace_root)? {
```
epage: What is your use case for `--allow-dirty`?

If its because you are using the dirty-tree code-gen pattern, personally I recommend against it.  It just isn't worth giving up the safety checks of denying dirty trees imo.  I recommend [codegenrs](https://github.com/crate-ci/codegenrs) pattern instead.

However, I don't think that is reason enough to block this going through.  My primary reason for asking is because I think we need more nuance to `git::is_dirty`.  I know of use cases to release/publish with untracked files but I am not aware of any with tracked files.  Tracked files are a lot more likely to trip up users because `cargo-release` will (most likely) commit on the user's behalf and we probably shouldn't be including "stray files" in that commit, especially when releasing workspaces where the "stray files" get associated with the first crate to commit.

So I think we should have a `allow_untracked: bool` (or `enum`) parameter for `git::is_dirty`.

src/main.rs
```diff
@@ -683,12 +683,17 @@ struct ReleaseOpt {
...
+    allow_dirty: bool,
+
+
     #[structopt(long)]
     /// Skip release confirmation and version preview
     no_confirm: bool,
 
     #[structopt(long)]
-    /// The name of tag for the previous release.
+    /// The name of tag for the previous release
```
epage: Have a thing against `.`? :)

(seeing you chose to fix our inconsistency by removing them rather than by adding them)

ArminBecher: Not at all. 
I just like it to be consistent. ;) 
So either a dot or no dot at the end?

src/main.rs
ArminBecher: I am with you that it is not a recommended default behavior to allow a dirty git repository for releases (that's why the flag should be optional and be set by the user explicitly).

However, I often find myself  in the situation that I am working on a small project on my own without feature branches and just commit my stuff to master. While this might not the most elegant and best way of using git (especially for larger projects), I strongly believe that I am not alone. Now it sometimes happens that I do some experiments and have a dirty working space, but still want to release project as it is without the changes. Without an `allow-dirty flag` I would have to stash my local changes in order to release.

The naming decision for `--allow-dirty`  comes from other cargo commands such as `cargo publish` which [has an `allow-dirty` flag](https://doc.rust-lang.org/cargo/commands/cargo-publish.html). By the way, I also pass the flag to `cargo publish`, if the --allow-dirty` flag is set.

Does this make sense?

src/main.rs
epage: No preference, just thought it interesting.

src/main.rs
epage: > Now it sometimes happens that I do some experiments and have a dirty working space, but still want to release project as it is without the changes. Without an allow-dirty flag I would have to stash my local changes in order to release.

So you have in-work changes but you want to release `HEAD`?  I would have assumed `cargo publish` would upload your working tree and not `HEAD`.  Even if it does, this will run into problems with the fact that `cargo-release` is not selective in what changes it commits.  We could make it more selective but then people might get in the habit of having dirty working trees until one day one of the dirty files happens to be one of the ones `cargo-release` changes.

ArminBecher: > Does this make sense?

No it doesn't make too much, to answer my own question. Thanks for your reply. You are right, `cargo publish` uploads the working tree and not `HEAD`, which makes my feature wish useless.

PR #181, opened 2019-11-09
Allow commits in a workspace to be consolidated.

epage:
To support workspace-wide configuration, we had to make a small tweak to how the config layering works.  We already needed to do this to help people who want different defaults for their workspace and their root crate.  I thought it already behaved this way and was suggesting people take advantage of it to work around their problems.

I was lazy and also included a documentation update for a past change.


epage:
> How about make consolidation a command line only option? With which we can avoid conflicts between workspace and crate level configuration.

a command line argument would be annoying to pass in every time.  We could use the workspace-config concept for this though.  When I get a chance, I'll give that a try.

PR #187, opened 2019-12-01
Give users better control over file-replacements

### LINKED ISSUE #113: Option to enforce replacements
It would be nice to have a option to enforce replacements, and fail otherwise.

Like:

```
{ file = "..", search = "...", replace = "..", min_matches = 1 }
```

### LINKED ISSUE #185: Option to exclude file replacements for pre-release?
If I'm going to switch to my changelog being updated by cargo-release and I also switch to using pre-release, this will end up being important

epage:
This has two breaking changes
- For each replacement, an error will occur if no matches for the pattern are found.  This can be reverted with `min = 0`
- Replacements will only happen on major/minor/patch releases. To have a replacement happen on a pre-release, set `prerelease = true`.

Fixes #185 
Fixes #113

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -64,10 +64,10 @@ bumping version:
 ```toml
 pre-release-replacements = [
   {file="CHANGELOG.md", search="Unreleased", replace="{{version}}"},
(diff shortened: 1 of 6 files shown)


sunng87:
I'm afraid it's our travis configuration causing the timeout error. I will merge this and adjust it later on.

PR #191, opened 2019-12-04
Feature/bump to specific version

### LINKED ISSUE #188: Bump to an exact release version
Example use cases:
- Align on a version across all crates in a workspace
- Bumping major with a pre-release after introducing breaking changes.  Currently you can only bump major or pre-release, not both.

sunng87:
Fixes #188 

This patch added support for bumping version to a fixed one. And the given version has to be:

* A valid semver string
* A version that is greater than current one

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,6 +16,7 @@ use std::ffi::OsStr;
 use std::io::Write;
 use std::path::Path;
 use std::process::exit;
+use std::str::FromStr;
 
 use boolinator::Boolinator;
 use chrono::prelude::Local;
@@ -176,13 +177,15 @@ impl<'m> PackageRelease<'m> {
             template.render(config.tag_name())
         };
 
+        let mut is_pre_release = false;
         let version = {
             let mut potential_version = prev_version.version.clone();
-            if let Ok(bump_level) = BumpLevel::from_str(&args.level) {
+            if let Ok(bump_level) = version::BumpLevel::from_str(&args.level) {
                 // bump level
                 if bump_level.bump_version(&mut potential_version, args.metadata.as_ref())? {
                     let version = potential_version;
                     let version_string = version.to_string();
+                    is_pre_release = bump_level.is_pre_release();
                     Some(Version {
                         version,
                         version_string,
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -767,12 +786,8 @@ struct ReleaseOpt {
...
 
     /// Release level: bumping specified version field or remove prerelease extensions by default
-    #[structopt(
-        possible_values(&version::BumpLevel::variants()),
-        case_insensitive(true),
-        default_value = "release"
-    )]
-    level: version::BumpLevel,
+    #[structopt(case_insensitive(true), default_value = "release")]
+    level: String,
```
epage: What about renaming this to `level_or_version`?  iirc that will make the intent clearer in `--help`

src/main.rs
```diff
@@ -767,12 +786,8 @@ struct ReleaseOpt {
     workspace: clap_cargo::Workspace,
 
     /// Release level: bumping specified version field or remove prerelease extensions by default
```
epage: Does the `--help` text (via the doc) need updating?

src/main.rs
```diff
@@ -767,12 +786,8 @@ struct ReleaseOpt {
...
-    #[structopt(
-        possible_values(&version::BumpLevel::variants()),
-        case_insensitive(true),
-        default_value = "release"
-    )]
-    level: version::BumpLevel,
+    #[structopt(case_insensitive(true), default_value = "release")]
```
epage: I believe we could set this up to be an `enum` of either a String or a Level.  That would push the error into arg parsing which might be clearer.

Also, can we still specify variants?  Is that actually used for validation or only for `--help`?

sunng87:
Thanks @epage for your great suggestion! I will update this later today.

src/main.rs
sunng87: The variants is for validation so I have to remove it. I wish clap has support for a default field in an arg_enum but unfortunately it doesn't. I will check its support for enum type.

epage: Oh well.

sunng87: I found enum support for structopt has other meaning, which is typically as subcommand. I will leave this as it is for now.

PR #193, opened 2019-12-11
 feat(change): Allow excluding paths with globs

### LINKED ISSUE #149: "no crate change" detection claims a crate is changed when the file isn't relevant
A crate might use `package.exclude` to not have documentation or other files included in the crate but if those files change, `cargo-release` will think the crate needs to be published.

Extra publishes shouldn't hurt and the use case is relatively small, I'm assuming.

epage:
We merge the crate's `excludes` with our own.  We aren't supporting `includes` out of simplicity.

Currently, this does not let you override the built-in excludes.  We can
change that (though that would be a breaking change) based on feedback.

Fixes #149


release.toml
```diff
@@ -5,5 +5,7 @@ pre-release-replacements = [
   {file="README.md", search="Current release: [a-z0-9\\.-]+", replace="Current release: {{version}}"} ,
   {file ="Cargo.toml", search="branch=\"[a-z0-9\\.-]+\"", replace="branch=\"{{version}}\""},
   {file="CHANGELOG.md", search="Unreleased", replace="{{version}}"},
-  {file="CHANGELOG.md", search="ReleaseDate", replace="{{date}}"}
+  {file="CHANGELOG.md", search="ReleaseDate", replace="{{date}}"},
+  {file="CHANGELOG.md", search="<!-- next-header -->", replace="<!-- next-header -->\n\n## [Unreleased] - ReleaseDate"},
```
sunng87: :+1:

PR #196, opened 2019-12-19
195: Allow Dirty

dbcfd:
Provide option to ignore untracked files when determining if working directory
is dirty.

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -7,7 +7,7 @@ use bstr::ByteSlice;
 use crate::cmd::call_on_path;
 use crate::error::FatalError;
 
-pub fn is_dirty(dir: &Path, ignore_untracked: bool) -> Result<bool, FatalError> {
+pub fn is_dirty(dir: &Path) -> Result<bool, FatalError> {
     log::debug!("Checking if path {:?} is dirty", dir);
     let output = Command::new("git")
         .arg("diff")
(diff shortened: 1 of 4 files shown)


src/git.rs
```diff
@@ -18,15 +19,26 @@ pub fn is_dirty(dir: &Path) -> Result<bool, FatalError> {
...
-    let output = Command::new("git")
-        .arg("ls-files")
-        .arg("--exclude-standard")
-        .arg("--others")
-        .current_dir(dir)
-        .output()
-        .map_err(FatalError::from)?;
-    let untracked_files = String::from_utf8_lossy(&output.stdout);
-    let untracked = !untracked_files.as_ref().trim().is_empty();
+    log::trace!("Unclean={}", tracked_unclean);
```
epage: Should this (and other path-traces) be traces at a higher, like a warn or error?

For `cargo publish`, the userfacing message includes this information and I think it'd help with debugging

src/main.rs
```diff
@@ -920,6 +920,10 @@ struct ConfigArgs {
     /// Do not create dev version after release
     no_dev_version: bool,
 
+    #[structopt(long)]
+    /// Ignore untracked files when determining whether working directory is dirty
+    ignore_untracked: bool,
```
epage: I think instead of having a flag to work around a bug, we should focus on the bug

dbcfd: I'm not entirely sure it is a bug in 0.13, or just different behavior and untracked wasn't taken into account before.

This is definitely useful for CI/CD environments, where you may have untracked files in the current directory due to CI/CD processes.

epage: It was added because `cargo-publish` complains about them.

Either I misunderstood that or else I'd align with `cargo-publish` and have `--allow-dirty`.

dbcfd: Will switch to allow dirty. Should satisfy my use case as well.

dbcfd: Done in https://github.com/sunng87/cargo-release/pull/196/commits/ed5cc454f18bd8f1cb990724480b666552671f8c

src/git.rs
dbcfd: I've moved them to info, since in most cases there should be no untracked files, and the list should be empty. Let me know if you want them at warn.

epage: Oh right, if there is nothing to log, we should skip logging.

src/main.rs
```diff
@@ -317,9 +317,12 @@ fn release_workspace(args: &ReleaseOpt) -> Result<i32, error::FatalError> {
     };
 
     git::git_version()?;
-    if git::is_dirty(&ws_meta.workspace_root, args.config.ignore_untracked)? {
-        log::warn!("Uncommitted changes detected, please commit before release.");
-        if !args.dry_run {
+    if git::is_dirty(&ws_meta.workspace_root)? {
+        if !args.config.allow_dirty {
```
epage: We should also pass `--allow-dirty` to cargo-publish

epage:
Been thinking about this some more and a bit hesitant to move forward with it.  For now, instead of focusing on a PR, could you create an issue that includes the use case for why you want to allow dirty so we can make sure we are covering the higher level need and it aligns with the rest of cargo-release?

PR #197, opened 2019-12-20
Add `--token` argument

Jake-Shadle:
Currently, the `cargo publish` step uses whichever token is in the users `~/.cargo/credentials`, or can be overriden by `CARGO_REGISTRY_TOKEN` (for crates.io), but it is sometimes convenient to specify the `--token` parameter for cargo publish, which this PR adds.


sunng87:
LGTM

PR #207, opened 2020-03-23
Increase timeout of wait_for_publish

dalance:
The timeout of `wait_for_publish` may be short because it requires disk access.
If the host is under heavy I/O load, `wait_for_publish` takes more than 30s.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -687,7 +687,7 @@ fn release_packages<'m>(
             )? {
                 return Ok(103);
             }
-            let timeout = std::time::Duration::from_secs(30);
+            let timeout = std::time::Duration::from_secs(300);
 
             if pkg.config.registry().is_none() {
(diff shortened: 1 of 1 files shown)


CreepySkeleton:
I think 5 minutes is too much in most cases.

Maybe also tweak the logic to have two different timeouts:
* Global timeout.
* One try timeout: the minimum amount of time `cargo publish` should wait between attempts. 

If an attempt has failed, repeat until the overall amount of time exceeds the global timeout.

dalance:
This timeout is used at only crates.io's server trouble (ex. synchronization failure between crates.io API server and GitHub).
Other usual troubles like user's network disconnected are failed immediately.
So I think too long timeout is not problem.

sunng87:
I agree a relatively large timeout could reduce affect from network, temporary latecy in crates.io. 

Merged.

epage:
As a point of input on this, I hit the timeout last week.  Thanks for doing tbis!

PR #226, opened 2020-08-29
Better error message when file to replace not found

meain:
Previously the error message just said file not found as it was just `IOError`.
Added custom logic to check and show the filename in case a file needed for replace is not found.


meain:
Possible fix to https://github.com/sunng87/cargo-release/issues/157

epage:
Thanks!

PR #227, opened 2020-09-27
Upgrade crates-index->0.15

nickolay:
This updates git2-rs to 0.13, which [supports MacPorts users better](https://github.com/rust-lang/git2-rs/issues/180#issuecomment-678740147).


nickolay:
We previously pinned rustc version to 1.37.0. This older compiler can't build the newer libraries, so I had to update travis.yml to the newest compiler

sunng87:
Thanks 👍

nickolay:
Wow, that was fast!

Any chance you could release a new version of cargo-release with this change? This would make `cargo install cargo-release` work for macports users out-of-box...

sunng87:
@nickolay hello! This has benn released in 0.13.7

nickolay:
Awesome, thank you!

PR #229, opened 2020-10-10
Add levels increment alphaMinor and alphaMajor 

librelois:
It is frequent to go through alpha pre-releases to test a new minor or major version.

If I want to start the test campaign for the `2.0.0` version of my project currently in `1.4.2` I have to specify by hand the name of the new version with its alpha suffix, it's a pity.

The `alphaMajor` level solves this problem.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -20,9 +20,11 @@
   * Otherwise, bump the patch field (0.1.0 -> 0.1.1)
 * `minor`: Bump minor version (0.1.0-pre -> 0.2.0)
 * `major`: Bump major version (0.1.0-pre -> 1.0.0)
-* `alpha`, `beta`, and `rc`: Add/increment pre-release to your version
+* `alpha`, `beta`, and `rc`: Increment pre-release to your version, or add pre-release **and** bump the patch field  
   (1.0.0 -> 1.0.1-rc.1, 1.0.1-dev -> 1.0.1-rc.1, 1.0.1-rc.1 ->
   1.0.1-rc.2)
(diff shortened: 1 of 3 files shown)


epage:
I feel like combining options like this is startin to get a bit unwieldy.  I wonder if instead we should have the option to take an absolute version  as an alternative to relative versions.

This would also help with workspaces where at one-off times crates get releases but you also want a synchronized version occasionally.

librelois:
> I feel like combining options like this is startin to get a bit unwieldy.

Do you see an easier way to publish a 1st alpha in a project without dev version (which is my case) ?

It is also possible to hide these advanced levels of help as long as it remains simple (those who want to dig deeper always have the reference).

>  I wonder if instead we should have the option to take an absolute version as an alternative to relative versions.

I'm not sure I understand. It's already possible to write a version by hand so I guess you're talking about something else, what do you mean by "an absolute version"?

epage:
> Do you see an easier way to publish a 1st alpha in a project without dev version (which is my case) ?

What I'm proposing is that instead of 
```
cargo publish alphaMinor
```
which I feel is complicating things and doesn't scale (would we also have an rcMinor or a  betaMinor?), I'm wondering if its better to instead do
```
cargo publish 1.1-alpha.1
```
We'd want to have a check to make sure the version is higher than the crates we are bumping but this gives the user control to do what they want at the cost of losing the convenience of the relative helpers (like `minrr`).

sunng87:
I agree with @epage that this case is already supported via absolute version.  I use absolute version bump frequently in my workflow.  It was reported in #188 and implemented in #191. It's more flexible and keep the complexity at a low level: imagine someone wants betaMinor or betaMajor.

Anyway, thanks for your patch @librelois

PR #233, opened 2020-10-21
(fix) set default logging level to debug when dry-run

### LINKED ISSUE #232: No --dry-run output by default - update README?
I've ran `cargo release --dry-run` as per the README, only to get a very terse output:

```
$ cargo release --dry-run
[2020-10-20T17:48:42Z INFO ] Update mycrate to version 0.2.6
[2020-10-20T17:48:42Z INFO ] Starting mycrate's next development iteration 0.2.7-alpha.0
[2020-10-20T17:48:42Z INFO ] Pushing HEAD to origin
[2020-10-20T17:48:42Z INFO ] Pushing v0.2.6 to origin
```

I recognized that this is formatted just like `env_logger`, which I use in my crate, so the next thing I tried was `RUST_LOG=trace cargo release --dry-run`, only to get the same output.

So, I looked up issues in this repo related to `--dry-run`, and finally found https://github.com/sunng87/cargo-release/issues/161 - where I discovered that I need to use `cargo release --dry-run -vv` which does output the commands as the README suggests.

This is kind of hard to discover though, especially as not even https://github.com/sunng87/cargo-release/blob/master/docs/reference.md notes that `cargo release` supports the `-v` flag. I am not too experienced with Rust so this might be a common pattern, but I would like to suggest - would it be possible to mention this in the README for newbies like me? :)

Thanks!

sunng87:
Fixes #232

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -904,7 +904,7 @@ impl Verbosity {
     /// Get the log level.
     pub fn log_level(&self, dry_run: bool) -> log::Level {
         let default_level = if dry_run {
-            3 // DEBUG if dry_run
+            4 // DEBUG if dry_run
         } else {
             2 // INFO
         };
--- a/src/main.rs
+++ b/src/main.rs
@@ -902,8 +902,13 @@ pub struct Verbosity {
 
 impl Verbosity {
     /// Get the log level.
-    pub fn log_level(&self) -> log::Level {
-        let verbosity = 2 - self.quiet + self.verbose;
+    pub fn log_level(&self, dry_run: bool) -> log::Level {
+        let default_level = if dry_run {
+            3 // DEBUG if dry_run
+        } else {
+            2 // INFO
+        };
+        let verbosity = default_level - self.quiet + self.verbose;
 
         match verbosity {
             std::i8::MIN..=0 => log::Level::Error,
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -1078,7 +1083,7 @@ pub fn get_logging(level: log::Level) -> env_logger::Builder {
 fn main() {
     let Command::Release(ref release_matches) = Command::from_args();
 
-    let mut builder = get_logging(release_matches.logging.log_level());
+    let mut builder = get_logging(release_matches.logging.log_level(release_matches.dry_run));
```
epage, line 1086: I feel this would be surprising to users to see different output on a regular run compared to doing a dry-run.

sunng87: Just verified again we will have to set default level to `TRACE` to see those commands we will execute. 

The original idea of dry-run is to show the use commands instead of actually to run it. So I think it's ok to have these additional lines. The only confusing part might be the different behaviour for `level` option

epage: I disagree, a dry-run is just that, a run without side effects to help people try things out first.  People expect the same behavior with dry-run as when they then remove it.

What people want to verify before running is different between each person.  For some, it might just be the default view which shows version numbers.  Others might want more.

As a first time user, if I saw output from dry-run and then ran again and did not see that output, I would worry something is going wrong.

So overall, I disagree with this approach for trying to help raise awareness of this logged output.  With that said, it is your project.

sunng87: I understand your concern. Adding implicit differences between dry-run and normal execution is bringing confusion. 

Updating README seems to be a better solution for #232

PR #237, opened 2020-11-24
fix(replace): Show a single diff in dry-run

epage:
It is difficult to test out replacements because each replacement gets
its own Diff, you have to mentally piece it all together.

With this change, we batch up the changes so all of them appear in a
single diff.


sunng87 (approved):
LGTM

PR #240, opened 2020-12-12
fix(reference): correct a small typo

woodruffw:
Noticed this while reading the reference. Thanks a ton for this amazing tool, and for making it so configurable!


epage:
Thanks for helping make it just that much better!

epage (approved):

sunng87:
Thanks!

PR #244, opened 2020-12-17
chore: Upgrade small dep breakages

epage:
Things left out
- semver is coupled to cargo-metadata which is coupled which is cupled
  to clap-cargo
- semver-parser completely changed its API, being higher level than
  before which makes our job harder to make tweaks to existing version
  ranges.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,9 +26,9 @@ checksum = "eab1c04a571841102f5345a8fc0f6bb3d31c315dec879b5c6e42e40ce7ffa34e"
 
 [[package]]
 name = "assert_fs"
-version = "0.13.1"
+version = "1.0.0"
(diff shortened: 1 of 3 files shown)


sunng87:
We can stick with current version of semver-parser. 

By the way, judging from your line widths, are you using any tool to create this PR?

sunng87 (approved):

epage:
> By the way, judging from your line widths, are you using any tool to create this PR?

No, I was doing it by hand.  I would recommend adopting Dependabot though.  I have it on all of my projects.  I have it set to update dependencies once a month.  It creates a PR per top-level dependency for updating either `Cargo.lock` or that and `Cargo.toml`.  For changes that don't break us, it makes it convenient to keep up to date.  When a dependency does break us, it at least raises awareness.

PR #246, opened 2020-12-21
add version into consolidated commit message

### LINKED ISSUE #222: When enable consolidate-commits, commit message's placeholder not been replaced
release.toml

```toml
disable-push = true
disable-publish = true
disable-tag = true
consolidate-commits = true
pre-release-commit-message = "Release {{version}}"
post-release-commit-message = "Released {{version}}, starting {{next_version}}"
```
I try to add pre-release-commit-message template config or use the default message template.

commit message:
```
Author: jolestar <jolestar@gmail.com>
Date:   Tue Jul 21 15:41:24 2020 +0800

    Released {{version}}, starting {{next_version}}

commit e7ce20ec0473f75f4275dbd99582494b6f4f2369
Author: jolestar <jolestar@gmail.com>
Date:   Tue Jul 21 15:40:36 2020 +0800

    Release {{version}}

```

pguedes:
only if all packages have the same version

Fixes #222

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -665,6 +665,13 @@ fn release_packages<'m>(
         let shared_commit_msg = {
             let template = Template {
                 date: Some(NOW.as_str()),
+                version: pkgs
+                    .iter()
+                    .map(|p| p.version.as_ref().map(|v| v.version_string.as_str()))
+                    .fold(
+                        pkgs[0].version.as_ref().map(|v| v.version_string.as_str()),
+                        |acc, v| if acc == v { acc } else { None },
+                    ),
                 ..Default::default()
             };
             template.render(ws_config.pre_release_commit_message())


src/main.rs
```diff
@@ -665,6 +665,13 @@ fn release_packages<'m>(
         let shared_commit_msg = {
             let template = Template {
                 date: Some(NOW.as_str()),
+                version: pkgs
+                    .iter()
+                    .map(|p| p.version.as_ref().map(|v| v.version_string.as_str()))
+                    .fold(
+                        pkgs[0].version.as_ref().map(|v| v.version_string.as_str()),
+                        |acc, v| if acc == v { acc } else { None }
+                    ),
```
epage, line 674: I'm torn.  This is an easy, low configuration way of getting a common version.

On the other hand, what is the likelihood that someone will rely on this and not notice when it falls apart on them because they were maintaining same-version by accident and not by design?

If we are concerned about that, I see two options
- A config field for lock-step versioning (error on mismatch, bump to ceiling (maybe with CLi flag?).  We could then safely set the version in this template
- Supporting absolute version bumps (`cargo release 2.0.0`) in addition to relative (`cargo release minor`).  This would be a recover mode in the error case and would help people go from `1.2.0.alpha5` to `2.0.0.alpha5` which we currently don't support.

@sunng87 thoughts on whether to go the implicit or explicit route?

pguedes: @epage yep i also initially thought of a cli flag that would fail if not all packages had the same version... but then was not super confident that would be the correct thing so i mostly wanted to start a discussion to figure out how this could work. (im happy to see it picked up so quickly :clap: )

this is pretty much the simplest solution i could think of to fix my personal use case, but i would be surprised if that covered all others :)

also, i'm not familiar with all functionality of cargo release (like absolute version bumps) so i could really use some :eyes: on this to make sure i contribute something useful to the project... that said i do have some time over the holidays since ill be stuck at home and would be happy to contribute

pguedes: bumpity?

sunng87: Sorry for late response.

@epage 's Option 2 is already supported in current release.

I'm afraid the solution in this patch was not easy to explain to user. With multi-versioned workspace, the version  `None` might be unexpected. 

Instead, I would suggest to make all the crate names and versions available to the template. Also perhaps a special version field with absolute version bump in welcomed.

kpcyrd: As a user, I would prefer having "Version could be None in some cases" as a documented caveat, right now I'm considering migrating away from cargo-release again back to custom scripts because my fairly basic and common use-case can't be implemented with cargo-release as of today. :)

hwchen: @sunng87 can you clarify when you say that @epage's option 2 is already supported? I tried with an absolute version bump but the commit message template didn't pick it up (still getting back `release {{version}}`). I might have made a mistake though or misunderstood.

sunng87: @hwchen Option 2 is supported when calling `cargo release 2.0.0`. But `version` is not available in consolidated commit message at the moment.

clux: Maybe an option here is to add an optional `{{ common_version }}` to the template context? That way people have to actually commit to depending on the same-version setup.

epage: #279 will make it so we could have an `if args.level_or_version.is_absolute()` check, if we wanted.

Another thought I had with this is if we had a flag for "unified version".  It would be set per-crate (but layering means it can inherit a global value) and we would find the highest of all "unified versions" and then use that as the base for relative version changes.   If this is set we would then provide the version in commit messages / tags.

By finding the max, it means a hot fix for a single crate within a workspace could be released and, on the next release, all of the other crates would be bumped to match it.

epage: The last thought I had was to introduce a variable for the root version.  If the workspace has a crate in the root, we would put that version in for variable substitution.  This is assuming that the root crate represents the whole.  We'd need to figure the naming of this.  `root_version`?

epage:
I'm assuming our including the version when `shared-version = true` supersedes this.

Issue #224, opened 2020-08-04
Wait between releases of workspace crates/Retry on failure


sebschmi:
I just released a workspace, and cargo release failed to verify a tarball. The tarball was valid though, it was just that one of its dependencies was uploaded to crates.io only seconds before, which was apparently not enough for crates.io to process the dependency. Running publish on the same crate again after a few more seconds worked totally fine.

The same effect occurred when doing a release completely manually without cargo release. Again, waiting a few seconds between releasing the two crates that depend on each other fixed the problem.

Therefore I propose to not abort if a tarball fails to verify, but instead to wait a few seconds and try again.

epage:
Could you clarify what you mean by "tarball failed to verify"?  There isn't a step related to that in cargo-release.

There is an issue where crates.io asynchronously publishes crates, so you can't `cargo publish` two crates in a row but instead need to ensure the first crate is fully published.  To workaround this, we [poll crates.io until the crate is published](https://github.com/sunng87/cargo-release/blob/master/src/cargo.rs#L59) or [300 seconds have passed](https://github.com/sunng87/cargo-release/blob/master/src/main.rs#L701).

sebschmi:
Sure, though sadly now I do not have the log anymore.

What basically happened was that I was running cargo release on a workspace, and it published the first crate just fine.
Then it published a second crate that has a dependency on the first crate, but there cargo publish failed, stating that the tarball failed to verify because the first crate does not exist.

Note that I ran cargo release half manually, by first changing all the versions to the correct strings. In detail, all my crates in my workspace were set to version 0.1.0, and all dependency declarations of local crates were also set to that version. So everything matched up.
I then ran cargo release release, and the behavior described above occurred.
________________________________

nickbabcock:
I'm trying to work through this issue as well. [The crate in question.](https://github.com/rakaly/jomini)

command:

```
cargo release patch --workspace
```

output (fails with "error: failed to prepare local package for uploading")

```
Release
  jomini_derive 0.1.2
  jomini 0.1.2
? [y/N]
y                                                                                                                                                                                                                [2020-08-06T23:45:11Z INFO ] Update jomini_derive to version 0.1.2                                                                                                                                               [master 3f13cbf] Release 0.1.2
 1 file changed, 1 insertion(+), 1 deletion(-)
[2020-08-06T23:45:11Z INFO ] Update jomini to version 0.1.2
[master 00e7c0a] Release 0.1.2
[...]

PR #247, opened 2020-12-26
add small thread sleep between publishes - for #224

clux:
Hey all. Thank you for cargo-release. It's very helpful!
This particular issue (#224) keeps biting me, and while this PR has only a dumb/ugly solution, I do think it should fix the problem. If the chosen time is bad, we could parametrise the second value into argv.

(it's probably possible to write a more advanced solution with a curl to crates.io to check that the previous upload has completed, but I'm not even sure that's perfect if all that's happening under the hood is that the write operation (the publish) propagates to all of crates' replicas.)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -702,8 +702,8 @@ fn release_packages<'m>(
 
             if pkg.config.registry().is_none() {
                 cargo::wait_for_publish(crate_name, &base.version_string, timeout, dry_run)?;
-                // give crates a little leeway to propagate a successful publish
-                // this should avoid failed to select failures where the next pkg in line depends on the previous
+                // HACK: Even once the index is updated, there seems to be another step before the publish is fully ready.
+                // We don't have a way yet to check for that, so waiting for now in hopes everything is ready
                 std::time::Duration::from_secs(5);
             } else {
                 log::debug!("Not waiting for publish because the registry is not crates.io and doesn't get updated automatically");
--- a/src/main.rs
+++ b/src/main.rs
@@ -702,6 +702,9 @@ fn release_packages<'m>(
 
             if pkg.config.registry().is_none() {
                 cargo::wait_for_publish(crate_name, &base.version_string, timeout, dry_run)?;
+                // give crates a little leeway to propagate a successful publish
+                // this should avoid failed to select failures where the next pkg in line depends on the previous
+                std::time::Duration::from_secs(5);
             } else {
                 log::debug!("Not waiting for publish because the registry is not crates.io and doesn't get updated automatically");
             }


src/main.rs
```diff
@@ -702,6 +702,9 @@ fn release_packages<'m>(
 
             if pkg.config.registry().is_none() {
                 cargo::wait_for_publish(crate_name, &base.version_string, timeout, dry_run)?;
```
epage, line 704: > (it's probably possible to write a more advanced solution with a curl to crates.io to check that the previous upload has completed, but I'm not even sure that's perfect if all that's happening under the hood is that the write operation (the publish) propagates to all of crates' replicas.)

In theory, that is what `wait_for_publish` is doing but for some  reason, it isn't.  Looking at people's logs in #224, we are waiting, for example
```
[2020-09-10T07:57:34Z INFO ] Waiting for publish to complete...
[2020-09-10T07:57:40Z INFO ] Running cargo publish on kube
```
Here we waited for 6s.  Yet somehow, it wasn't good enough.  Apparently, the part of the index we are checking gets updated but not everything needed to publishing.  We are then racing for that to be updated.

src/main.rs
```diff
@@ -702,6 +702,9 @@ fn release_packages<'m>(
...
+                // give crates a little leeway to propagate a successful publish
+                // this should avoid failed to select failures where the next pkg in line depends on the previous
+                std::time::Duration::from_secs(5);
```
epage, line 707: Maybe update the comment, based on my above analysis

```
// HACK: Even once the index is updated, there seems to be another step before the publish is fully ready.  We don't have a way yet to check for that, so waiting for now in hopes everything is ready
```
Also, should this be in `wait_for_publish`?  On one hand, it keeps all related logic together.  On the other hand, its helpful to have hacks stand out.

epage (changes requested):
We need to reach out to the crates.io team to figure out what is happening but that shouldn't block making things work for people.

src/main.rs
clux: Yeah, I am not sure. AFAIKT from the documentation in https://docs.rs/crates-index/0.16.2/crates_index/struct.Index.html you are doing the right thing in `wait_for_publish`.

src/main.rs
clux: Have updated the comment. I think the `wait_for_publish` is doing the right thing, so maybe it just takes some times for writes to crates.io to propagate to all its read-replicas. At any rate, it's only a problem when publishing more than one crate, so probably good to keep the hack here.

epage: The insiduous thing about race conditions is you can never know if your change actually helped because you might have just not hit the race when testing your fix.  Or the problem might go away and you don't need the hack anymore

:(

epage (approved):

yrashk:
It looks like I am hitting the issue of not waiting enough before one of the newly published crates is available before proceeding to the next one. Makes it much harder to resume the process.

yrashk:
Maybe it would make sense to allow continuing if trying to re-upload a crate with the same version?

PR #251, opened 2021-01-17
Push tag prior to full push

### LINKED ISSUE #250: Push tag before branch
Currently, cargo-release pushes the branch first, then the tag. When you have a CI that creates artifacts and attaches them to the tagged release, having the branch pushed first delays CI starting because the branch push has to be processed first, delaying the availability of the artifacts.

sunng87:
Fixes #250 

I just simply moved tag push code above the branch push. However, I'm not quite sure if we can optimize the `pushed` check in the new case. And, will the tag push contain all the necessary commits?


epage (approved):
Sorry, somehow missed this

PR #256, opened 2021-03-22
fix sleep after publish (call sleep) for #224;

greenhat:
Ref #224
Seems like in #247 a call to `sleep` was missing.


epage:
Surprised we all missed this.  Wish it wasn't needed though.

sunng87:
:joy:

Issue #218, opened 2020-07-01
How can I get this to publish from CircleCI? Failure on timeout for publishing crate (appears related to using SSH for git)


DarthHater:
Hi there!

This crate is really cool, and I'm stoked to get it working in my workflow.

That said, I'm getting an issue when trying to release using CircleCI:

https://app.circleci.com/pipelines/github/sonatype-nexus-community/cargo-pants/19/workflows/3d29f4b9-8907-421c-9a8f-67a70ffb5b70/jobs/20

```
    Updating crates.io index
 Downloading crates ...
  Downloaded cargo-release v0.13.4
  Installing cargo-release v0.13.4
 Downloading crates ...
  Downloaded dirs v2.0.2
  Downloaded dirs-sys v0.3.5
  Downloaded humantime v1.3.0
  Downloaded home v0.5.3
  Downloaded libssh2-sys v0.2.18
  Downloaded cfg-if v0.1.10
  Downloaded env_logger v0.7.1
  Downloaded difference v2.0.0
  Downloaded combine v3.8.1
  Downloaded jobserver v0.1.21
  Downloaded glob v0.3.0
  Downloaded aho-corasick v0.7.13
  Downloaded miniz_oxide v0.3.7
  Downloaded rustc-demangle v0.1.16
  Downloaded fnv v1.0.7
  Downloaded syn-mid v0.5.0
  Downloaded walkdir v2.3.1
  Downloaded version_check v0.9.2
  Downloaded linked-hash-map v0.5.3
  Downloaded adler32 v1.1.0
  Downloaded quick-error v1.2.3
  Downloaded unicode-width v0.1.8
  Downloaded thread_local v1.0.1
  Downloaded quote v1.0.7
  Downloaded itoa v0.4.6
  Downloaded lazy_static v1.4.0
  Downloaded pkg-config v0.3.17
  Downloaded proc-macro2 v1.0.18
  Downloaded num-traits v0.2.12
  Downloaded unreachable v1.0.0
  Downloaded termcolor v1.1.0
  Downloaded structopt v0.3.15
  Downloaded ryu v1.0.5
  Downloaded percent-encoding v2.1.0
  Downloaded textwrap v0.11.0
  Downloaded serde v1.0.114
  Downloaded unicode-xid v0.2.1
  Downloaded memchr v2.3.3
  Downloaded serde_json v1.0.56
  Downloaded serde_derive v1.0.114
  Downloaded same-file v1.0.6
[...]

PR #257, opened 2021-04-19
Allow to set the time waited between two cargo publishs

sebschmi:
This adds a command line parameter `--publish-grace-sleep` that defines how long we wait between two invocations of `cargo publish`. The default value is 5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@
 
 ### Added
 
-* Add `--publish-grace-sleep` parameter that allows to set the number of seconds to sleep between
-  two invocations of `cargo publish`
+* Add `PUBLISH_GRACE_SLEEP` environment variable that allows to set the number of seconds to sleep between
+  two invocations of `cargo publish`. Default is `5`
 * Do not sleep between publishes on dry runs
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+* Add `--publish-grace-sleep` parameter that allows to set the number of seconds to sleep between
+  two invocations of `cargo publish`
+* Do not sleep between publishes on dry runs
+
 ### Changed
 
 * New `diable-release` config flag to skip crates in a workspace
--- a/src/main.rs
+++ b/src/main.rs
@@ -710,6 +710,7 @@ fn release_packages<'m>(
                 // HACK: Even once the index is updated, there seems to be another step before the publish is fully ready.
                 // We don't have a way yet to check for that, so waiting for now in hopes everything is ready
                 if !dry_run {
+                    log::info!("Waiting an additional {} seconds for crates.io to update its indices...", args.config.publish_grace_sleep);
                     std::thread::sleep(std::time::Duration::from_secs(args.config.publish_grace_sleep));
                 }
             } else {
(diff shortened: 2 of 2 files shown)


epage:
Sorry no one has responded on this.  This could actually help in #218.

There is a part of me that would prefer for this to be an env variable, like `RUST_BACKTRACE`.   I think part of it stems from this being a bit of a stopgap feature and not wanting to bake it into the CLI and clutter up the help.

@sunng87 or @sebschmi any thoughts on that?

sebschmi:
I would be totally fine with an env variable as well.

Has btw anyone made an issue at https://github.com/rust-lang/crates.io/issues about some API feature to poll if a crate is ready?

epage:
> Has btw anyone made an issue at https://github.com/rust-lang/crates.io/issues about some API feature to poll if a crate is ready?

My understanding is that git is the API; we've just not been able to determine what we are doing differently than `cargo` when it determines a crate is available.

sebschmi:
I made an [issue at crates.io](https://github.com/rust-lang/crates.io/issues/3677) now anyways. I think it would be useful to have a simple way to automate crate publishing without complicated hacks.

sunng87:
@epage Do you have any suggestion about the criteria to classify the location of configuration, among cli options, toml files or environment variables? 

I think I have a little sense with the difference between cli options and toml files. But for environment variable, I can't tell it from cli options.

sunng87:
@sebschmi you might want to format the code to pass our rustfmt check.

epage:
> @epage Do you have any suggestion about the criteria to classify the location of configuration, among cli options, toml files or environment variables?
> 
> I think I have a little sense with the difference between cli options and toml files. But for environment variable, I can't tell it from cli options.

Im honestly going off of gut feel here and can be completely wrong :)

I think `typos` is where I;ve started to most formalize my thoughts on CLI inputs:
- User / integration relevant settings (e.g. output format): I'm using the CLI at the moment because there are few and defaults are good enough.  I might add user config later.
- Repo relevant settings these live in a config file in the repo
- Iterative repo relevant settings: I exposed these both on the CLI and in the repo's config file (e.g. excluding files from being analyzed)

So far, I only use env variables to comply with the various CLI colored output standards.

In this case, the wait time doesn't really fit into any of the categories.  This is a workaround for another bug and there isn't anything about this that is repo -specific.  For user-relevant, it sort of is just by how much the user can tolerate failure vs delays.  For integration-relevant, it sort of is to help us with #218 where there is another bug/limitation in our wait logic.

So based on my earlier classifications. this most closely aligns with how I've been using CLI args / user-config.  The reason I'm considering other approaches, like modeling after some crate processes, like `RUST_BACKTRACE`, is because this is a workaround.  `cargo-release` does not have a user config file and it already has a sizeable CLI "API" for people to navigate.  I don't see this being something that should have raised visibility and being as formally specified.  This led me to the thought of env variables.

sebschmi:
@sunng87  I will keep out of the discussion of this being a CLI argument or env variable. But as soon as you came to an agreement, I will make the necessary changes, also to make it pass the CI checks.

sunng87:
> because this is a workaround

@epage sounds good to me. I agree with the idea of putting it into environment variables. cc @sebschmi

sebschmi:
I changed the code, and kept the impact of the env variable minimal: I simply directly query it at the `// HACK` comment.

I did not document this feature, if you want to I can do it.

sunng87:
@sebschmi sorry for late response. I will update document for you.

sebschmi:
No worries, thank you very much for merging!

Issue #263, opened 2021-06-15
Uncommited changes uncorrectly determines workspace root


TotalKrill:
I have a project that is a workspace, but the each crate in the workspace is its own git submodule. 

This makes cargo release refuse to update the submodule crate, before i commit the new submodulde changes to the root workspace. leaving me with the error message: 
![image](https://user-images.githubusercontent.com/3248006/122064495-ba7c2a80-cde0-11eb-8bd0-d9341b2b38d4.png)

I assume this is not correct, and I should be able to do a release of a submodule crate without having to update the entire workspace git repo.

epage:
(sorry, never explicitly posted)

We have not kept sub-module support in mind when implementing various features.  Some are low hanging and I have one PR up (and another soon) to address them.

The last two are
- Dirty repo check
- "Behind push-remote" check

Maybe we have a "check one" vs "check all" based on `conslidate-commits`

epage:
At this point, we only have an advisory message not updated for submodules.  I would be curious what your experience is like now.

TotalKrill:
It has actually been kinda horrible, it took me quite some time to realize where the problem was.

Each developer is basically developing an own feature or software in our system. We then use the entire system or workspace as a way to consolidate how each of these separate functions or crate are fitting into the entirety.

 This way it feels like we are not loosing velocity. The main issue before I managed to identify where the problem was. Is that the message we got did not indicate that we should be looking in any directory above each developers own feature or crate.

So each had assumed there was a bug, and as we have tests and some features that depend on that the git tag, and the crate version to consolidate the entire system. Not being able to use cargo release has resulted in some annoying tag and crate version mismatches, and this bugs :)

We now have some internal tools to check this. But just knowing that it was a workspace issue would have helped us :)

epage:
Sorry, my intention was to ask how cargo-release does with sub-modules after making changes to better work with it, to see if there are use cases we aren't yet covering, etc.

kornelski:
I'm still getting "ERROR cargo_release::release] Uncommitted changes detected, please commit before release." in the latest version.

Is there a way to tell cargo-release *not* to check git at all?

btatting:
[...]

PR #265, opened 2021-06-16
feat(git): Warn when releasing from an odd git state

### LINKED ISSUE #264: Help users from releasing a local divergence from push-remote's branch
Several times, I forgot that I merged some PRs from others and forgot to `git pull` before doing a `cargo release`.

Options
- Before confirmation prompt, warn or error if the local branch is behind the remote branch
- Automatically pull w/ rebase against remote's branch

I'm unsure how universally safe it is to do things against the remote, so the options range from advisory to automatic.

epage:
I'm starting with a conservative approach, just warning.  We'll see if
anyone complains.  If they do, we can keep it as-is.  If they don't, we
can explore turning this into a hard error.

Fixes #264

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -494,7 +494,21 @@ fn release_packages<'m>(
 
     // STEP 0: Help the user make the right decisions.
     let lock_path = ws_meta.workspace_root.join("Cargo.lock");
+    let mut pulled: HashSet<_> = HashSet::new();
     for pkg in pkgs {
+        let git_remote = pkg.config.push_remote();
+        if !pulled.contains(git_remote) {
+            let branch = git::current_branch(&ws_meta.workspace_root)?;
+            if branch == "HEAD" {
+                log::warn!("Releasing from a detached HEAD");
+            }
+            git::fetch(&ws_meta.workspace_root, git_remote, &branch)?;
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -494,7 +494,21 @@ fn release_packages<'m>(
 
     // STEP 0: Help the user make the right decisions.
     let lock_path = ws_meta.workspace_root.join("Cargo.lock");
+    let mut pulled: HashSet<_> = HashSet::new();
     for pkg in pkgs {
+        let git_remote = pkg.config.push_remote();
```
sunng87, line 499: Perhaps we can lift this above the for-loop? Git is a workspace-scoped concept and we do not need to check it for each module.

epage: We just currently allow a remote per package, see below in the `push` section.  I mirrored that pattern here for consistency.  Despite being in the loop, at least we will, in practice, only fetch once since people will likely only have one remote.

The code this mirrored I think predates us having workspace-wide config.  That was introduced with `consolidate_commits`.  We could look into moving the remotes to being workspace-wide but I think that should be a separate PR because that is technically a breaking change, despite the fact that probably no one will be broken.

Granted, we do have #263 for sub-module support.  No idea what that should look like and how any of our existing features need to change for it, if we support it.

Issue #266, opened 2021-06-17
Support git push --push-option


imp:
`git push` allows for generic parameters to be added to pass arbitrary strings to the servers. It is done via `--push-option/-o` flag. It is used by gitlab for triggering and manipulating CI (and perhaps used by other systems as well). It would be very useful to allow setting these parameters via `release.toml` or flags.
I would imaging syntax like this:
```
push-options = ["text1", "text2", "key=value"]
```

Which will generate the following git command
```
git push --push-option text1 --push-option text2 --push-option key=value ...
```

epage:
Could you look at #268 to see if it will work for your needs?

imp:
That's perfect! Thanks so much! Any chance to release it to crates.io soon?

sunng87:
A new release 0.15.0 is created for this

PR #267, opened 2021-06-18
fix(git)!: Dealing with the remote is workspace-wide

epage:
No idea how this will interact with #263 but this will help with
implementing #266.

BREAKING CHANGE: When in a workspace, we'll now only look at the
workspace's config for `disable-push` and `push-remote` and no the
package-specific config.


sunng87 (approved):

sunng87:
LGTM

PR #272, opened 2021-06-22
Fix Target Specific Dependencies

### LINKED ISSUE #271: Target-Specific Workspace Dependency Not Bumped
When one member of the workspace has a target-specific dependency on another member of the workspace, that dependency is not properly bumped. This can be easily reproduced using the following repository: https://github.com/cwfitzgerald/cargo-release-target-dependency-bug. In this example `user` has a non-wasm dependency on `library`.

When you run `cargo release --skip-publish --skip-push --skip-tag --no-confirm` on the repo, you'll get the following error:

```
[2021-06-21T22:39:46Z INFO ] Starting library's next development iteration 0.1.1-alpha.0
[2021-06-21T22:39:46Z WARN ] Fatal: Invalid TOML file format: Error during execution of `cargo metadata`: error: no matching package named `library` found
    location searched: C:\Users\conno\Programming\cargo-release-target-dependency-bug\library
    prerelease package needs to be specified explicitly
    library = { version = "0.1.1-alpha.0" }
    required by package `user v0.1.0 (C:\Users\conno\Programming\cargo-release-target-dependency-bug\user)`
```

cargo-release seems not to be looking where it needs to to replace these dependencies. 

I am using the current version on crates.io: 0.15.0

cwfitzgerald:
Fixes #271. Tested and works correctly on both the test case in the issue and the original repo that caused the problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Fixed
+* Fixed issue where the versions of cfg specific dependencies wouldn't be properly bumped.
+
 ## [0.15.0] - 2021-06-19
 
 ### Added
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -143,6 +143,32 @@ pub fn set_dependency_version(
             }
         }
 
+        if manifest.as_table().contains_table("target") {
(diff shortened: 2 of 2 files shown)


sunng87 (approved):

sunng87:
Thanks!

PR #279, opened 2021-07-05
refactor(args): Abstract level vs version args

epage:
Besides separating concerns, this is a step towards a potential solution
to #222, see discussion in #246.  Now we are preserving whether the user
used an absolute or relative version.


sunng87 (approved):

sunng87:
Sorry for late. The patch looks good to me

PR #281, opened 2021-07-15
fix(git): respect 'disable_push' flag at package level

lucab:
This introduces an additional check for the 'disable_push' option
at the package level, to allow specific setting for a single entry.
The corresponding workspace option is still evaluated before this.

Closes: https://github.com/sunng87/cargo-release/issues/280


epage (approved):

epage:
Thanks!

PR #285, opened 2021-07-31
fix: Don't treat bump as an exclude

### LINKED ISSUE #248: Update semver, clap-cargo, and cargo_metadata
These are coupled to each other and need to be updated in lock step.  I suspect a new release of `clap-cargo` is needed.

### LINKED ISSUE #249: Update semver-parser
When I tried to do this, I gave up.  We use `semver-parser` as a way to programmatically modify version numbers but they've made it into an intermediate-level crate that does some interpretation.  It'd be a lot of work to transform back into the user's version number.

### LINKED ISSUE #284: cargo release "exclude" option prevent semver incrementation
Hi,

I'm using `cargo-release` on a project with the following architecture

- root crate
  - utilities: local only crate
  - python: python bindings

When I run `cargo-release`, I would like to increment the version number of both the root crate and the python subcrate, while excluding the utilities crate.

I'm using the following command line:
~~~
cargo release --dry-run --no-dev-version --skip-publish --workspace --exclude utilities patch
~~~
But I noticed that the semver wasn't incremented at all:
~~~
[2021-07-29T14:16:34Z INFO ] Pushing python-v0.1.0 to origin
[2021-07-29T14:16:34Z INFO ] Pushing HEAD to origin
[2021-07-29T14:16:34Z INFO ] Pushing v0.1.0 to origin
[2021-07-29T14:16:34Z INFO ] Pushing HEAD to origin
~~~

Only when you removed the `--exclude utilities` option, the incrementation works as expected:
~~~
cargo release --dry-run --no-dev-version --skip-publish --workspace patch
[2021-07-29T14:17:26Z INFO ] Update python to version 0.1.1
[2021-07-29T14:17:26Z INFO ] Update utilities to version 0.1.1
[2021-07-29T14:17:26Z INFO ] Update bug_release to version 0.1.1
[2021-07-29T14:17:26Z INFO ] Pushing python-v0.1.1 to origin
[2021-07-29T14:17:26Z INFO ] Pushing HEAD to origin
[2021-07-29T14:17:26Z INFO ] Pushing utilities-v0.1.1 to origin
[2021-07-29T14:17:26Z INFO ] Pushing HEAD to origin
[2021-07-29T14:17:26Z INFO ] Pushing v0.1.1 to origin
[2021-07-29T14:17:26Z INFO ] Pushing HEAD to origin
~~~

I created a repository for you to reproduce the issue:
https://github.com/Wenzel/bug_cargo_release

Is this a bug, or is there something that I'm missing ?

Thanks !

epage:
Running
`cargo release --dry-run --no-dev-version --skip-publish --workspace --exclude utilities patch`

Treated `patch` as an `exclude` because of a naive use of the `clap`
crate.  This updates to the version of `clap_cargo` with the fix as well
as a whole slew of dependencies.

`clap_cargo`, `cargo_metadata`, `semver`, and `semver_parser` are in
lock-step with each other.  Thankfully, `semver` 1.0 is more like
the old `semver_parser` (before the merge `semver_parser` gained some
`semver` policy) so we could actually upgrade.  This did require
re-implementing some old `semver` logic because it was dropped.

The annoying part was that `cargo_metadata` migrated from `std::path` to `camino`, requiring extra type conversions.

Fixes #284
Fixes #249
Fixes #248

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,13 +11,12 @@ readme = "README.md"
 edition = "2018"
 
 [dependencies]
-cargo_metadata = "0.9"
-crates-index = "0.16"
+cargo_metadata = "0.14"
+crates-index = "0.17"
 toml = {version = "0.5", default-features = false}
 toml_edit = "0.2.1"
 serde = { version = "1.0", features = ["derive"] }
-semver = "0.9.0"
-semver-parser = "0.9.0"
+semver = "1.0"
 quick-error = "2.0"
 regex = "1.0"
 bstr = "0.2.8"
@@ -32,7 +31,7 @@ itertools = "0.9"
 once_cell = "1.2.0"
 structopt = {version = "0.3.0", default-features = false}
 clap = { version = "2", default-features = false }
-clap-cargo = { version = "0.3", features = ["cargo_metadata"] }
+clap-cargo = { version = "0.5", features = ["cargo_metadata"] }
 log = "0.4"
 env_logger = "0.8"
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -32,6 +32,6 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions-rs/toolchain@v1
       with:
-        toolchain: 1.46
+        toolchain: 1.52
         override: true
     - run: cargo build --all-features
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.15"
+version = "0.7.18"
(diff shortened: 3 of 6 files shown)


Cargo.toml
```diff
@@ -11,13 +11,13 @@ readme = "README.md"
 edition = "2018"
 
 [dependencies]
-cargo_metadata = "0.9"
-crates-index = "0.16"
+cargo_metadata = "0.14"
+camino = "1.0"
```
sunng87, line 15: camino is not used in our code directly so we better skip this to avoid conflict

epage: Good catch.  At first, I expected to have to reference it directly but ended up being able to avoid it.  Forgot to remove it.

epage: Updated

sunng87 (approved):

PR #296, opened 2021-08-12
(feat) New flag --exclude-unchanged

tonyg:
In order to publish just those crates in a workspace that have changed (plus their transitive dependencies), I've added an `--exclude-unchanged` flag. It further filters the `pkgs` selection resulting from consulting the workspace and the `--package` and `--exclude` flags.


epage:
Thanks for getting the ball rolling on this.  Could you open an issue about this as a place to discuss how we should expose this and how it would work?  I feel like it'd be better to capture that conversation over there (e.g. if this PR gets supplanted by another one, unlikely but still) and not to clog up the code discussion with the design discussion.

PR #321, opened 2021-08-25
fix: next_version not rendered in post commit message

sunng87:
By turning off consolidate-commits for this project. A better approach
is to provide additional tempalte variables to consolidate-commits
template if current project is not a workspace, or has a fixed release
version.


epage:
Missed this since I was unifying config with all other projects but left post-release bump which I don't use elsewhere

Issue #339, opened 2021-10-08
cargo release --execute times out everytime


rbtcollins:
E.g. `[2021-10-08T14:03:20Z ERROR cargo_release] Fatal: Timeout waiting for crate to be published.`

What is happening there and how can we get more insight to troubleshoot/fix it?

rbtcollins:
Oh and the crates were https://crates.io/crates/unleash-api-client and https://crates.io/crates/rocket-enumform

yamadapc:
Logs print:
```
[2021-10-30T06:25:41Z DEBUG cargo_release::cargo] Crate index update failed with invalid version 0 on git_proxy_options; class=Invalid (3)
```

Until timeout.

yamadapc:
Upgrading `libgit2` has fixed the issue for me.

```
brew upgrade libgit2
```

epage:
I'll try vendoring libgit2 to see if that helps.

PR #349, opened 2021-10-30
Update to crates-index@0.18

yamadapc:
Updates crates-index and its usage.

I thought this was related to https://github.com/crate-ci/cargo-release/issues/339 (it is not really).

A dependency bump is probably ok here though. Replaced `.retrieve` with `.update` as retrieve seems to be deprecated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -200,12 +200,11 @@ dependencies = [
 
 [[package]]
 name = "crates-index"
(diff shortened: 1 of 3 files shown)


epage (approved):
Thanks! I hadn't realized they had released the new version.  Eventually dependabot would have told us but still nice to be ahead.

PR #360, opened 2021-11-28
Support for tab completion

woodruffw:
First of all, thanks for creating `cargo-release`! I use it in virtually all of my Rust projects, and it's an absolute pleasure to use.

This PR introduces self-generating tab completion (aka "shell completion" in some shells) to `cargo-release`. It uses `clap`'s built-in tab completion generation, by way of `structop`. Every shell completion dialect supported by `clap` is transparently supported by `cargo-release`.

By way of example, this command can be used to load shell completions for `cargo-release` into a `bash` session:

```bash
eval "$(cargo-release completions --shell=bash)"
```

This is similar but not identical to the CLI provided by tools like `rustup`, which exposes `rustup completions [tool] [shell]` to provide completions.

Let me know what you think! I'm happy to tweak the behavior, as desired.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -15,6 +15,7 @@ pub enum Command {
         setting = structopt::clap::AppSettings::DontCollapseArgsInUsage
     )]
     Release(ReleaseOpt),
+    Completions(CompletionsOpt),
 }
 
 #[derive(Debug, StructOpt)]
@@ -221,4 +222,10 @@ fn resolve_bool_arg(yes: bool, no: bool) -> Option<bool> {
         (false, false) => None,
         (_, _) => unreachable!("StructOpt should make this impossible"),
     }
+}
+
+#[derive(StructOpt, Debug, Clone)]
+pub struct CompletionsOpt {
+    #[structopt(long, short = "s", possible_values = &clap::Shell::variants(), case_insensitive = true)]
+    pub shell: clap::Shell,
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,7 +4,7 @@
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::ffi::OsStr;
-use std::io::Write;
+use std::io::{self, Write};
 use std::path::Path;
 use std::process::exit;
 
@@ -27,16 +27,25 @@ use crate::replace::{do_file_replacements, Template};
 use crate::version::VersionExt;
 
 fn main() {
-    let args::Command::Release(ref release_matches) = args::Command::from_args();
+    match args::Command::from_args() {
+        args::Command::Release(ref release_matches) => {
+            let mut builder = get_logging(release_matches.logging.log_level());
+            builder.init();
 
-    let mut builder = get_logging(release_matches.logging.log_level());
-    builder.init();
-
-    match release_workspace(release_matches) {
-        Ok(code) => exit(code),
-        Err(e) => {
-            log::error!("Fatal: {}", e);
-            exit(128);
+            match release_workspace(release_matches) {
+                Ok(code) => exit(code),
+                Err(e) => {
+                    log::error!("Fatal: {}", e);
+                    exit(128);
+                }
+            }
+        }
+        args::Command::Completions(ref completion_matches) => {
+            args::Command::clap().gen_completions_to(
+                "cargo-release",
+                completion_matches.shell,
+                &mut io::stdout(),
+            );
         }
     }
 }


woodruffw:
The "Lint Commits" action is failing, although I'm not sure whether it's my fault or not :slightly_smiling_face: 

```
Linting commits:
* 099d328 main, args: support for tab completion
Against 'committed.toml':
ignore_author_re = 'dependabot'
subject_length = 50
subject_capitalized = true
subject_not_punctuated = true
imperative_subject = true
no_fixup = true
no_wip = true
hard_line_length = 0
line_length = 72
style = 'conventional'
allowed_types = [
    'fix',
    'feat',
    'chore',
    'docs',
    'style',
    'refactor',
    'perf',
    'test',
]
merge_commit = false

If this fails, don't sweat it. We're trying to encourage clear communication and not hinder contributions.
If it is a reasonable issue and you lack time or feel uncomfortable fixing it yourself,
let us know and we can mentor or fix it.
099d3286839be9569c62455a85549effd46b3a80: error Invalid commit format invalid commit format
```

src/args.rs
```diff
@@ -222,3 +223,9 @@ fn resolve_bool_arg(yes: bool, no: bool) -> Option<bool> {
         (_, _) => unreachable!("StructOpt should make this impossible"),
     }
 }
+
+#[derive(StructOpt, Debug, Clone)]
+pub struct CompletionsOpt {
+    #[structopt(long, short = "s", possible_values = &clap::Shell::variants(), case_insensitive = true)]
```
epage, line 229: Any particular reason you did case insensitive?

src/args.rs
```diff
@@ -15,6 +15,7 @@ pub enum Command {
         setting = structopt::clap::AppSettings::DontCollapseArgsInUsage
     )]
     Release(ReleaseOpt),
+    Completions(CompletionsOpt),
```
epage, line 18: How come you did it at this level?  I'm assuming most people will be running `cargo-release` as `cargo release` and not `cargo-release`.

epage:
> The "Lint Commits" action is failing, although I'm not sure whether it's my fault or not slightly_smiling_face

The complaint is

> 099d3286839be9569c62455a85549effd46b3a80: error Invalid commit format invalid commit format

ie it isn't following [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) but also note it says 

> If this fails, don't sweat it. We're trying to encourage clear communication and not hinder contributions.
> If it is a reasonable issue and you lack time or feel uncomfortable fixing it yourself,
> let us know and we can mentor or fix it.

I need to go back and look at getting color enabled (the underlying tool supports it) so it'll be easier to see the problem.

src/main.rs
```diff
@@ -27,16 +27,25 @@ use crate::replace::{do_file_replacements, Template};
...
+                    exit(128);
+                }
+            }
+        }
+        args::Command::Completions(ref completion_matches) => {
+            args::Command::clap().gen_completions_to(
+                "cargo-release",
+                completion_matches.shell,
+                &mut io::stdout(),
+            );
```
epage, line 48: How should we document this capability, including giving the example code you had in the R?

We could probably put it in the long help for the command.

src/args.rs
woodruffw: Nope, no particular reason. If I had to contrive one, people might do `--shell=Bash` or similar and reasonably expect that to work. But I don't mind changing it to case sensitive at all, to keep the possibilities simpler.

src/args.rs
woodruffw: I put it at the top-level since that's how `rustup` and a few other tools in the Rust packaging/toolchain ecosystem do it. My argument for that would be that `cargo release` is how people should be running it interactively, but `cargo-release completions [...]` makes more sense in the context of a shell profile (to emphasize that the completions are coming from **just** `cargo-release` and that they aren't hooked up to the `cargo` completions yet).

src/main.rs
woodruffw: Hmm, perhaps an FAQ entry? The example I included in the top comment is extremely specific to `bash`, so including it in the long help might confuse users of other shells (zsh, fish, etc.)

We could also do it in the "reference" docs, although those seem to cover just the `release` subcommand so perhaps not...

woodruffw:
> ie it isn't following [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) but also note it says

Thanks! I misread that error as the tool complaining about the commit's *hash*, not the commit message body. I'll tweak and amend the PR to bring it into compliance.

PR #369, opened 2021-12-23
Don't try to bump versions of excluded packages

benesch:
Suppose a workspace containing:

   * Crate A, v1
   * Crate B, v2

A command like `cargo release -p crate-a 1.1.0` would fail because
`cargo release` would attempt to bump crate B to v1.1.0 while loading it
(which is not legal). This goes unnoticed when running something like
`cargo release -p crate-a minor` because crate B's version gets bumped
to v2.1.0 (and then this in-memory version bump is harmlessly
discarded).

The solution here is not the most elegant, but it works.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -79,7 +79,7 @@ fn release_workspace(args: &args::ReleaseOpt) -> Result<i32, error::FatalError> 
     let root = git::top_level(ws_meta.workspace_root.as_std_path())?;
     let pkg_releases: Result<HashMap<_, _>, _> = selected_pkgs
         .iter()
-        .filter_map(|p| PackageRelease::load(args, &root, &ws_meta, p).transpose())
+        .filter_map(|p| PackageRelease::load(args, &root, &ws_meta, p, true).transpose())
         .map(|p| p.map(|p| (&p.meta.id, p)))
         .collect();
     let pkg_releases = pkg_releases?;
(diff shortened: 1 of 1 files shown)


epage:
Please open an issue and let's try to understand the problem and how it should be fixed.

I've run similar cases to that many times and with verbose logging and I've not seen anything suspect.  This doesn't mean there isn't a bug but that more understanding is needed before deciding how to fix it.

benesch:
Ok, I filed #370 and created a repository with a minimal reproduction: https://github.com/benesch/cargo-release-bug.

By the way, if you'd generally prefer issues before PRs, it'd be great to call that out in a CONTRIBUTOR.md or somesuch. For small issues that can easily be described in a PR description, I default to sending up the issue description and PR all in one. Personally I find it's often easiest to explain a bug by demonstrating a proof-of-concept fix, and it gives us something concrete to reference when describing possible fixes.

It's just always a bit disheartening to have a PR immediately closed, is all; it goes down much easier when the norms are spelled out in CONTRIBUTOR.md. :)

epage:
The problem with a PR-first approach is it focuses on **a** implementation rather than the problem and how to resolve it.  PR first can work when its trivial but even this PR admitted that there were problems with this approach.

PRs carry a weight to them, they are extra pressure on maintainers to keep them open because of what you mentioned.  Except for the more aloof maintainers, it takes a lot more mental and emotional energy to close out a PR like I did.  This isn't just isolated to me or this project but I've heard of it from others.  I'd recommend keeping that in mind generally.

benesch:
> The problem with a PR-first approach is it focuses on **a** implementation rather than the problem and how to resolve it. PR first can work when its trivial but even this PR admitted that there were problems with this approach.
> 
> PRs carry a weight to them, they are extra pressure on maintainers to keep them open because of what you mentioned. Except for the more aloof maintainers, it takes a lot more mental and emotional energy to close out a PR like I did. This isn't just isolated to me or this project but I've heard of it from others. I'd recommend keeping that in mind generally.

So, I have a completely different opinion on this for the projects I maintain! I personally love when someone shows up with a PR rather than an issue, because it demonstrates a increased level of expertise and investment in the project. Even if the PR has to be totally reworked, I'm always more excited to fix the bugs that come with PRs, since I know I have someone I can work _with_ to fix the bug.

It's totally reasonable for us to disagree here! Just wanted to call out the difference in norms, and that I do look to the CONTRIBUTOR.md to determine what the norms for a given project are.

PR #381, opened 2022-01-04
feat: Publish using `rust-version` toolchain

### LINKED ISSUE #380: Option to release using MSRV toolchain
This will be a last check to make sure things work with MSRV

Before releasing, we should check if all MSRV toolchains are available.

epage:
This is a last ditch effort to catch MSRV issues.

Fixes #380


docs/reference.md
```diff
@@ -52,11 +52,12 @@ Configuration is read from the following (in precedence order)
...
-| `push`         | `--no-push`   | bool   | Don't do git push |
+| `release`      | `--exclude`     | bool   | Skip the entire release process (usually for internal crates in a workspace) |
+| `push`         | `--no-push`     | bool   | Don't do git push |
 | `push-options` | \-              | list of strings | Flags to send to the server when doing a `git push` |
-| `tag`          | `--no-tag`    | bool   | Don't do git tag |
-| `publish`      | `--no-publish` |  bool | Don't do cargo publish right now, see [manifest `publish` field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish--field-optional) to permanently disable publish. |
+| `tag`          | `--no-tag`      | bool   | Don't do git tag |
+| `publish`      | `--no-publish`  |  bool | Don't do cargo publish right now, see [manifest `publish` field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish--field-optional) to permanently disable publish. |
+| `rust-version` |                 |  string | Toolchain version to use for publishing (specify `"manifest"` to use [Cargo.toml's `rust-version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) |
```
epage, line 60: Should we be concerned there isn't a way to turn this off?

I'm tempted to overlay this with a `bool` so you can just enable/disable it.  Just lazy about writing the serde logic for it.

src/cargo.rs
```diff
@@ -50,22 +50,62 @@ pub fn package_content(manifest_path: &Path) -> Result<Vec<std::path::PathBuf>,
...
+
+    let bin: String;
+    let formatted_toolchain_version;
+    if let Some(toolchain_version) = toolchain_version {
+        formatted_toolchain_version = format!("+{}", toolchain_version);
+        command = vec!["cargo", &formatted_toolchain_version];
+    } else {
+        bin = cargo();
+        command = vec![&bin];
+    }
```
epage, line 102: Should we be concerned that we are ignoring `CARGO` when using the `rust-version`?

PR #394, opened 2022-01-11
docs: correct description of CRATE_NAME in hook

jeremyBanks:


epage:
Thanks for fixing this!

PR #396, opened 2022-01-17
feat: Allow specifying build target

nicholasbishop:
Add `target` to the config (and `--target` argument) allowing the
verification build to be done against a different target triple. This is
helpful for crates that don't build with the default host target triple.


epage (approved):

epage:
v0.19.2 is now released.

PR #397, opened 2022-01-20
feat(cargo): Preserve dependent version format

nicholasbishop:
When updating a dependent version to a caret version, only include an
explicit caret if the existing version used one. This is implemented on
string types since `semver` doesn't distinguish between an implicit and
explicit caret version.

This is similar in spirit to the implementation in cargo-edit:
https://github.com/killercup/cargo-edit/commit/3de9db7329e847f456ed6cbbbe6ea6b57848f68b

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -229,9 +229,14 @@ fn find_dependency_tables<'r>(
     })
 }
 
-fn set_version(dep_item: &mut toml_edit::Item, name: &str, version: &str) -> bool {
+fn set_version(dep_item: &mut toml_edit::Item, name: &str, mut version: &str) -> bool {
     if let Some(table_like) = dep_item.as_table_like_mut() {
         if let Some(version_value) = table_like.get_mut("version") {
+            // Preserve the presence or lack of an explicit caret.
+            if version.starts_with('^') && !version_item_uses_caret(version_value) {
+                version = &version[1..];
+            }
+
             *version_value = toml_edit::value(version);
             true
         } else {
@@ -242,6 +247,14 @@ fn set_version(dep_item: &mut toml_edit::Item, name: &str, version: &str) -> boo
         log::debug!("Not updating version-only dependency on {}", name);
         false
     }
+}
+
+/// Check if a toml item representing a version starts with a caret.
+fn version_item_uses_caret(version: &toml_edit::Item) -> bool {
+    version
+        .as_str()
+        .map(|s| s.starts_with('^'))
+        .unwrap_or(false)
 }
 
 pub fn update_lock(manifest_path: &Path) -> Result<(), FatalError> {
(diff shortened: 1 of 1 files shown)


src/cargo.rs
```diff
@@ -229,9 +229,22 @@ fn find_dependency_tables<'r>(
...
+        .unwrap_or(false)
+}
+
+fn set_version(dep_item: &mut toml_edit::Item, name: &str, mut version: &str) -> bool {
     if let Some(table_like) = dep_item.as_table_like_mut() {
         if let Some(version_value) = table_like.get_mut("version") {
+            // Preserve the presence or lack of an explicit caret.
+            if version.starts_with('^') && !version_item_uses_caret(version_value) {
+                version = &version[1..];
+            }
```
epage, line 246: >  This is similar in spirit to the implementation in cargo-edit:

cargo-edit was based off of this (surprised I made that change between the two).  The big impediment to easily do this like like cargo-edit is that  our `set_requirement` takes a `VersionReq` while cargo-edit's `upgrade_requirement` takes  a `str`.

I don't remember why the two implementations diverged.  Ideally we'd make them more similar but this is fine for now.

src/cargo.rs
```diff
@@ -229,9 +229,22 @@ fn find_dependency_tables<'r>(
...
 }
 
-fn set_version(dep_item: &mut toml_edit::Item, name: &str, version: &str) -> bool {
+/// Check if a toml item representing a version starts with a caret.
+fn version_item_uses_caret(version: &toml_edit::Item) -> bool {
+    version
+        .as_str()
+        .map(|s| s.starts_with('^'))
+        .unwrap_or(false)
+}
```
epage, line 238: nit: Could you move this after `set_version`?  Since English is read top-to-bottom, that puts an emphais on what comes first, so I prefer to list the top-level item before its  internals

src/cargo.rs
```diff
@@ -728,6 +741,102 @@ mod test {
 
             temp.close().unwrap();
         }
+
+        /// Updating a dependent version that uses an explicit caret
+        /// should retain an explicit caret.
+        #[test]
+        fn preserve_caret() {
```
epage, line 748: Could you move this down to the test module?

epage:
Thanks for doing this!

src/cargo.rs
nicholasbishop: done

src/cargo.rs
nicholasbishop: I think it is already in the test module, unless I'm misreading. (Or do you mean move it from the nested `set_dependency_version` module to the parent `test` module?)

epage: Sorry, misread the collapsed diff

epage:
v0.19.3 is now released with this change

PR #402, opened 2022-01-30
fix(cli): Fix help text for --verbose

nicholasbishop:
The default verbosity is 2, which corresponds to info logging. Adding a
single `-v` adds debug logging, and a second one adds trace logging,
which is the most verbose level.


epage:
Thanks!

PR #411, opened 2022-02-01
fix: correct publish order for circular dev dependencies

### LINKED ISSUE #421: Wrong release order when using root crate
Looks like a duplicate of #366, but I'm still encountering the issue in version `0.20.1`. I'm using a [root package workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html#root-package)—[`Cargo.toml`](https://github.com/andybarron/depcon/blob/main/Cargo.toml) is the manifest for the main package _and_ specifies the workspace members.

## Repro steps
1. Clone [`andybarron/depcon`](https://github.com/andybarron/depcon)
2. Run `cargo release --workspace minor`
3. Observe that `cargo-release` attempts to publish root package `depcon` first, even though its updated dependency on `depcon_codegen@$NEW_VERSION` would prevent the publish from succeeding.

vmx:
This is a fix for the following case. You have a workspace with one root
package and another sub-package. That sub-package could e.g. be a derive
macro implementation. In order to test that derive macro, you need some
functionality from the root packge. Therefore the derive macro has a
dev-dependency on the root package.

When publishing such a workspace, currently the way the order is
determined, is by looking at the dependencies between the workspace
packages. In this case the root depends on the sub-package and the other
way round. As the packages are looked at an arbitrary order, it is
determined that the root package should be published first and then the
sub-package. This isn't the correct order though.

This commit fixes that problem by excluding dev and build dependencies,
when looking at the dependencies within the workspace.

Fixes #421

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -281,7 +281,19 @@ pub fn sort_workspace(ws_meta: &cargo_metadata::Metadata) -> Vec<&cargo_metadata
         .iter()
         .filter_map(|n| {
             if members.contains(&n.id) {
-                Some((&n.id, &n.dependencies))
+                // Return the package ID of all normal dependencies. Exclusde all dependencies that
+                // are dev and/or build dependencies only.
+                let normal_deps: Vec<_> = n
+                    .deps
+                    .iter()
+                    .filter_map(|d| {
+                        d.dep_kinds
+                            .iter()
+                            .position(|k| k.kind == cargo_metadata::DependencyKind::Normal)
+                            .map(|_| &d.pkg)
+                    })
+                    .collect();
+                Some((&n.id, normal_deps))
             } else {
                 None
             }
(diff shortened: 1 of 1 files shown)


src/cargo.rs
```diff
@@ -281,7 +281,19 @@ pub fn sort_workspace(ws_meta: &cargo_metadata::Metadata) -> Vec<&cargo_metadata
...
+                let normal_deps: Vec<_> = n
+                    .deps
+                    .iter()
+                    .filter_map(|d| {
+                        d.dep_kinds
+                            .iter()
+                            .position(|k| k.kind == cargo_metadata::DependencyKind::Normal)
+                            .map(|_| &d.pkg)
+                    })
+                    .collect();
```
epage, line 295: To be most correct, we should only ignore dev dependencies without the version field set.

To be mostly correct, we can ignore all dev dependencies.

We should still respect build dependencies though.

epage: (sorry for the delay in looking at this)

andybarron: Made this change in #422.

Unfortunately, it doesn't look like `cargo_metadata` can tell us whether a specific dependency version was actually requested, so we had to go with the "mostly correct" solution instead.

vmx: Thanks @andybarron for fixing this. I guess, we then keep going at #422 instead?

PR #422, opened 2022-02-15
fix: ignore dev dependencies in sort_workspace

### LINKED ISSUE #421: Wrong release order when using root crate

andybarron:
Same as #411 (H/T @vmx), but we only ignore dev dependencies; build dependencies are still respected.

Borrowing @vmx's excellent summary:

>This is a fix for the following case. You have a workspace with one root package and another sub-package. That sub-package could e.g. be a derive macro implementation. In order to test that derive macro, you need some functionality from the root packge. Therefore the derive macro has a dev-dependency on the root package.
>
>When publishing such a workspace, currently the way the order is determined, is by looking at the dependencies between the workspace packages. In this case the root depends on the sub-package and the other way round. As the packages are looked at an arbitrary order, it is determined that the root package should be published first and then the sub-package. This isn't the correct order though.
>
>This commit fixes that problem by excluding dev and build dependencies, when looking at the dependencies within the workspace.

Fixes #421.

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -281,7 +281,19 @@ pub fn sort_workspace(ws_meta: &cargo_metadata::Metadata) -> Vec<&cargo_metadata
         .iter()
         .filter_map(|n| {
             if members.contains(&n.id) {
-                Some((&n.id, &n.dependencies))
+                // Return the package ID of all normal dependencies. Exclusde all dependencies that
+                // are dev and/or build dependencies only.
+                let normal_deps: Vec<_> = n
+                    .deps
+                    .iter()
+                    .filter_map(|d| {
+                        d.dep_kinds
+                            .iter()
+                            .position(|k| k.kind == cargo_metadata::DependencyKind::Normal)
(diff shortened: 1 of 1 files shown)


epage (approved):

andybarron:
@epage fixed commit format for CI, need another approval for the workflows to run :pray:

epage:
v0.20.2 is now released

PR #431, opened 2022-03-07
fix dirty detection when a file is named `head`

### LINKED ISSUE #430: cargo-release incorrectly detects local changes if there's a file named `head` in root dir
In order to check for dirty files, cargo-release runs `git diff HEAD --exit-code --name-only`, checks the exit code, and splits the standard out. However, when there's a file named `head` in crate or workspace root, git fails with an exit code of `128` but no stdout output.  On stderr, it echoes this:

```bash
$ git diff HEAD --exit-code --name-only
fatal: ambiguous argument 'HEAD': both revision and filename
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

This then fails with the confusing verbose log:
```
[… DEBUG cargo_release::git] Dirty because of:

[… ERROR cargo_release::release] Uncommitted changes detected, please commit before release.
```
with no further information other than a blank line between those messages.

I'll throw together a quick PR with `--`, fixing this

jbr:
closes #430

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -59,6 +59,7 @@ pub fn is_dirty(dir: &Path) -> Result<bool, FatalError> {
         .arg("HEAD")
         .arg("--exit-code")
         .arg("--name-only")
+        .arg("--")
         .current_dir(dir)
         .output()
         .map_err(FatalError::from)?;


jbr:
closing in preference to #432

epage:
Looks like we raised to post a fix.  Since #430 has a little more to it, I'll stick with that.

epage:
And we race to close this PR ;)

Man, thought Rust was supposed to help us with these data races...

PR #437, opened 2022-03-29
chore: Delete and ignore fixtures' lock files

Person-93:
When I was browsing the codebase, my IDE was insisting on updating these lock-files, but I figured that it's not worth the bother of keeping them up to date, especially when you consider that the lock files might be different in the MSRV. Apparently the lock-file was changed in a recent release of Cargo.

Here's a PR with my suggested changes.


Person-93:
I got the tests to pass, but I'm not so confident in my recommendation any more.

src/cargo.rs
```diff
@@ -921,6 +923,8 @@ mod test {
 
         #[test]
         fn in_pure_workspace() {
+            update_lock(Path::new("tests/fixtures/pure_ws/Cargo.toml")).unwrap();
+
             let temp = assert_fs::TempDir::new().unwrap();
             temp.copy_from("tests/fixtures/pure_ws", &["**"]).unwrap();
```
epage, line 929: If we move forward with this, we should be updating the lock file in the temp dir rather than in the source

epage:
> but I'm not so confident in my recommendation any more.

What are you thinking now?

Person-93:
> > but I'm not so confident in my recommendation any more.
> 
> What are you thinking now?

I'm thinking that there may be other hidden gotchas that I'm unaware of. I'd defer to you because you know the codebase.

epage:
I lean towards simple and reproducible unless there is a reason otherwise.

PR #511, opened 2022-08-25
Add `--allow-dirty` flag

HCastano:
When publishing [`ink!`](https://github.com/paritytech/ink) we tend to remove `dev-dependencies` across our workspace before
publishing due to some cyclic dependencies. It would be nice to be able to remove them
temporarily and still be able to publish using `cargo-release`.

I'm not sure if there's any way to write tests for this, but if there is let me know and
I can add them.

Supersedes https://github.com/crate-ci/cargo-release/pull/196.

P.S cool tool!

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -2,16 +2,17 @@
 
 ## CLI Arguments
 
-| Argument        | Format | Description |
-|-----------------|--------|-------------|
-| `--execute`     | bool   | Actually perform the release |
+| Argument        | Format | Description                                                    |
+|-----------------|--------|----------------------------------------------------------------|
+| `--execute`     | bool   | Actually perform the release                                   |
 | `--no-confirm`  | bool   | Release the crate without the user verifying what will happen. |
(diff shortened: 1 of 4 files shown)


epage:
> When publishing [ink!](https://github.com/paritytech/ink) we tend to remove dev-dependencies across our workspace before
publishing due to some cyclic dependencies. It would be nice to be able to remove them
temporarily and still be able to publish using cargo-release.

Won't cargo-release then commit them, undoing any of that "temporary" nature of the change?

Is there a reason you include the version in the dev dependencies?  Dropping the version will cause `cargo publish` to remove the dependencies automatically on publish

HCastano:
> Won't cargo-release then commit them, undoing any of that "temporary" nature of the change?

We release using `--no-push`, so `cargo-release` never commits anything. 

> Is there a reason you include the version in the dev dependencies?

Not that I can remember, so I think we can get away with not specifying a version there. 

> Dropping the version will cause cargo publish to remove the dependencies automatically on publish

That's good to know!

PR #545, opened 2022-10-14
fix(publish): Bail out error for rate limits

### LINKED ISSUE #483: The tool attempts to release many crates too rapidly
```
error: failed to publish to registry at https://crates.io

Caused by:
  the remote server responded with an error (status 429 Too Many Requests): You have published too many crates in a short period of time. Please try again after Thu, 30 Jun 2022 15:07:44 GMT or email help@crates.io to have your limit increased.
```

epage:
Not thrilled with erroring but putting in a sleep for 10 minutes would probably make people unhappy.  We can always refine this further.

Fixes #483

### DIFF
--- a/src/steps/mod.rs
+++ b/src/steps/mod.rs
@@ -193,6 +193,57 @@ pub fn verify_monotonically_increasing(
     Ok(success)
 }
 
+pub fn verify_rate_limit(
+    pkgs: &[plan::PackageRelease],
+    index: &crates_index::Index,
+    dry_run: bool,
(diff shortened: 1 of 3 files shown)


rnbguy:
We have a cargo workspace of 31 packages. Would you accept a PR with dynamic rate limits (possibly via environment variable) instead of hard-coded `30` for existing crates?

epage:
I'd recommend creating an issue so we don't lose track of any design discussions

Issue #107, opened 2019-04-27
Find a better name for dependency-version config/flag


epage:
We added it in #101 but don't like the name.

epage:
Finding another model for this feature might be another way of making it easier to explain.

The goal is to update dependencies within the workspace.

Design ideas
- When version-bumping a crate, ensure dependents relationship to this crate is vaid
- Before publishing a crate, ensure all in-workspace dependencies are valid

epage:
Been looking into the concept of generating a changelog from fragments.  My thoughts are at https://github.com/epage/epage.github.io/issues/23#issuecomment-1750832190

I think we could make it so we get what dependents to bump from changelog fragments.

PR #548, opened 2022-10-15
[RFC] feat: add a way to always update the version in dependent Cargo.tomls

sunshowers:
In workspaces with independent versioning, it can be very useful to always keep dependencies on the latest version even if they're still compatible with the old version. Add a config `always-update-version` that, if set to true, always updates the version in dependent crates.

Looking for feedback on:

* [ ] Is this desirable for cargo-release? I know nextest needs this.
* [ ] What should the name of the option be? I just threw in a name as a placeholder.
* [ ] How should I test this? I noticed that `src/steps/version.rs` doesn't have any tests.

### DIFF
--- a/src/steps/version.rs
+++ b/src/steps/version.rs
@@ -279,7 +279,7 @@ pub fn update_dependent_versions(
                 }
             }
             config::DependentVersion::Fix => {
-                if !dep.req.matches(&version.bare_version) {
+                if pkg.config.always_update_version() || !dep.req.matches(&version.bare_version) {
                     let new_req =
                         crate::ops::version::set_requirement(&dep.req, &version.bare_version)?;
                     if let Some(new_req) = new_req {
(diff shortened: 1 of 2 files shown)


src/steps/version.rs
```diff
@@ -279,7 +279,7 @@ pub fn update_dependent_versions(
                 }
             }
             config::DependentVersion::Fix => {
```
epage, line 281: Why not enable `dependent-version = "upgrade"`?

Maybe I should make that the default for the next version...

epage: btw we've never really liked the name for this but haven't had better ideas.  See https://github.com/crate-ci/cargo-release/issues/107

sunshowers: Oh I didn't realize that was a config! Sorry, missed that :)

Issue #82, opened 2019-02-16
Pre-release versioning incompatible with [patch]


ryankurte:
Hey there,

I'm not sure whether this is strictly an _issue_, but it took me some time to work out and seems like a bit of a pain.

When `cargo release X` is run the version is bumped, the crate is published, then the version is again bumped to a development version. As cargo expects patch versions to match, this final bump means that you can no longer use `[patch]` overrides against the currently published library version for testing (without reverting the crate version in the patched crate's `Cargo.toml`).

This results in errors like:
```
warning: Patch `rr-mux v0.4.1-alpha.0 (/home/ryan/projects/rr-mux)` was not used in the crate graph.
```
Where the project `cargo.toml` contains:
```toml
[dependencies]
rr-mux = "0.4.0"
...

[patch.crates-io]
rr-mux = { path = "../rr-mux" }
```

One fix is to always run `cargo release` with `--no-dev-version` to avoid the problem. 

My suggestion would be that that the second development version bump should be _opt in_, I didn't expect it to happen, but, either way there's now some documentation for anyone else who has the same problem.

sunng87:
The last bump behaviour was learned from tools in Java and Python world. It indicates the current branch may have changes after last release.

The `--no-dev-version` is created for your case. And you can add it to `release.toml` to make it a default option for your current project.

ryankurte:
> The last bump behaviour was learned from tools in Java and Python world. It indicates the current branch may have changes after last release.

it _kindof_ makes sense in that regard, but it also seems incompatible with cargo. i personally would expect making a pre-release to also be an explicit command.

> The `--no-dev-version` is created for your case. And you can add it to `release.toml` to make it a default option for your current project.

yep, and i now use `--no-dev-version` every time. another possibility would be to add a global ~/.cargo/release.toml so it can be turned off completely.

as i said, i don't know whether to consider this _wrong_, it's just frustrating. i'm happy to open a PR altering the behaviour to be opt-in or documenting the incompatibility if one of those would be useful?

sunng87:
> yep, and i now use `--no-dev-version` every time. another possibility would be to add a global `~/.cargo/release.toml` so it can be turned off completely. 

Yes you can add it to `~/.release.toml` (and I think `~/.cargo/release.toml` is a better place for this) to disable this behaviour globally. 

[...]

PR #553, opened 2022-10-17
fix!: Remove dev-version support

epage:
This just adds enough complications to the workflow that I'm removing it.  People can always do this manually if they want with `cargo release version`.

- Makes it harder to `[patch]` dependencies (see #82)
- Gets in the way of detecting changes since last tag
- Incompatible with the `--unpublished` workflow in #551 
- Takes work to maintain this code without enough motivating of a case

See #82 for some details.


ojacobson:
A removal of this nature should probably have been treated as a breaking change, as release processes that used it are no longer compatible with it. Having that happen across a minor version upgrade was a surprise.

In the future, I'd also like to ask that workarounds and recommended alternatives be included in the changelog or elsewhere in the documentation, and not only in Github tickets.

epage:
This happened in 0.22 which is a breaking release in cargo's variant of semver.  In Cargo, you ignore leading `0.`s to determine what field acts like major.

ojacobson:
Oh, hell. My apologies. I understood this to be _one_-point-two-two, not zero. I'm out of line.

PR #585, opened 2022-10-31
Crates only count to the rate limit check when being published to crates.io

### LINKED ISSUE #584: Rate-limiting check doesn't respect "--no-publish"
When trying to publish the first version of a large workspace, I ran into the error message introduced in #545. 

My plan was to do a `cargo release --no-publish` to bump all the version numbers then publish manually to avoid rate limits, but it seems like `verify_rate_limit()` gets called even when you won't push to crates.io.

```console
$ cargo release 0.2.0-rc.1 --execute --no-publish
error: attempting to publish 18 new crates which is above the crates.io rate limit
```

Michael-F-Bryan:
Fixes #584. CC #545.

### DIFF
--- a/src/steps/mod.rs
+++ b/src/steps/mod.rs
@@ -220,7 +220,7 @@ pub fn verify_rate_limit(
     let mut new = 0;
     let mut existing = 0;
     for pkg in pkgs {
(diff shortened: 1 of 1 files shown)


epage:
Released as v0.22.2

PR #598, opened 2022-11-16
fix: Take the --unpublished option into account

alessandrod:
https://github.com/crate-ci/cargo-release/commit/11eab502414c3a0040466ac7fa7089e80a836eb1 added support for --unpublished but accidentally always turned it on


epage:
Released in v0.23.1

Issue #540, opened 2022-10-14
pre-release-commit-message template replacements don't seem to work


sunshowers:
Hi there!

I'm trying to adapt cargo-release to work for nextest. The nextest workspace consists of several crates, each with their own independent versioning.

If I'd like to update one of nextest's crates, I'd like to be able to do:

```
cargo release -p nextest-filtering 0.2.2 --execute
```

and have a commit be created with the requisite version bumps and the message:

```
[nextest-filtering] version 0.2.2
```

and a signed tag created on this commit called `nextest-filtering-0.2.2` with message:

```
nextest-filtering version 0.2.2
```

Here's my first attempt at configuring that, based on [the reference](https://github.com/crate-ci/cargo-release/blob/master/docs/reference.md#placeholders):

https://github.com/nextest-rs/nextest/blob/main/release.toml

However, when I run:

```
cargo release -p nextest-filtering 0.2.2 --execute
```

the commit that gets created is:

```
commit 49aaac68f9fae149b957ed79508e3551cd20d816 (HEAD -> main)
Author: Rain <rain@sunshowers.io>
Date:   Fri Oct 14 12:23:50 2022 -0700

    [{{crate_name}} version {{version}}
```

which indicates that template replacements aren't working.

Any idea how to make them work?

```
% cargo release --version
cargo-release 0.21.4
```

epage:
> consolidate-commits = true

This tells `cargo-release` to use a single commit for all version bumps which means there is not a specific crate name or version to use in case you run `cargo release --workspace` or `cargo release -p one -p two` (and we don't want it conditioned on how you run as that might cause problems).

Setting this back to `false` should get you exactly the behavior you want.

Note: in the next version, I'm looking at making `consolidate-commits = true` the default which is why I suggested still explicitly setting it.  The focus on the next release is to better scale up with large workspaces.

sunshowers:
Got it, thanks! Just managed to debug this myself as well.

sunshowers:
BTW I did a writeup of how nextest uses cargo-release: https://github.com/nextest-rs/nextest/blob/main/internal-docs/releasing.md

epage:
> For each changed crate, in topological order (internal dependencies like nextest-filtering first, cargo-nextest last), run cargo release -p <crate-name> <version>. For example, cargo release -p nextest-runner 0.28.0. This will perform a dry run.

In the upcoming version of `cargo-release`, my hope is that this will change to
[...]

PR #604, opened 2022-11-28
fix: Log when un-rendered template varabiables are present

epage:
Inspired by #540


MarijnS95:
Shouldn't this be logged as `warn`? Fwiw when I wrote https://github.com/crate-ci/cargo-release/issues/540#issuecomment-1328769105 I hoped this to break before pushing out the release, allowing users to rectify it.

epage:
Logging at debug is a first, conservative step, as we evaluate what should be happening.

PR #655, opened 2023-03-20
Enable feature "vendored-openssl" on dependency git2

NobodyXu:
so that `cargo-release` can build in environment where openssl isn't installed or hard to install (e.g. cross compilation) and the binaries built can be run on everywhere, without runtime dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ pre-release-replacements = [
 cargo_metadata = "0.15"
 crates-index = "0.19"
 # Enabling the feature for `crates-index`
-git2 = { version = "0.16.1", features = ["vendored-libgit2"], default-features = false }
+git2 = { version = "0.16.1", features = ["vendored-libgit2", "vendored-openssl"], default-features = false }
 toml_edit = "0.19.6"
 toml = "0.7.3"
 serde = { version = "1.0", features = ["derive"] }


Cargo.toml
```diff
@@ -32,7 +32,7 @@ pre-release-replacements = [
 cargo_metadata = "0.15"
 crates-index = "0.19"
 # Enabling the feature for `crates-index`
-git2 = { version = "0.16.1", features = ["vendored-libgit2"], default-features = false }
+git2 = { version = "0.16.1", features = ["vendored-libgit2", "vendored-openssl"], default-features = false }
```
epage, line 35: > so that cargo-release can build in environment where openssl isn't installed or hard to install (e.g. cross compilation) and the binaries built can be run on everywhere, without runtime dependencies.

What problem did you run into?

I believe the openssl-related features should be disabled because of `default-features = false`.

NobodyXu: crates-index enables git2/https by default, which in turns enable openssl-sys, openssl-probe and libgit2-sys/https.

libgit2-sys/https also pulls on openssl-sys.

epage: Could you add a comment, explaining that we are enabling it for `crates-index` so someone doesn't come along later and delete it, since it looks unused?

NobodyXu: I've added a comment for the additional of this new feature.

PR #675, opened 2023-05-09
docs: Changelog generation with git-cliff

joshka:
I understand the note about having good communication in changelogs, but for smaller projects maintaining this manually can be overkill. Using git-cliff avoids this extra tax of maintaining a changelog when you're the sort of person that does that in your commit history. The readme mentions hooks for this, but doesn't provide concrete guidance on how to set this up. This PR addresses that in the FAQ.

I hope I've maintained the original intent of the note about communication while changing it subtly so that a simple approach is also provided. Happy to reword if you'd prefer something closer to the original.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,10 +11,11 @@ Features
   - Optionally share commits
 - Handles `cargo publish`, tagging, and pushing
 - Pre-release search and replace for custom version updates, including
-  - Updating changelogs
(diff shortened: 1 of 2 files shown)


joshka:
Your changes LGTM. :)
My goal was to save others the minutes I took to experiment and get this setup simply, as well as documenting it in the right place for next time I need set it up.

Issue #692, opened 2023-08-04
Freeze with the sparse protocol


loyd:
I'm trying to release all crates in a workspace with the latest stable toolchain (1.71.1) and cargo-release (0.24.11) and after the `Published` step the whole process just freezes and nothing happens
```
note: Waiting for `elfo-dumper v0.2.0-alpha.5` to be available at registry `crates-io`.
You may press ctrl-c to skip waiting; the crate should be available shortly.
   Published elfo-dumper v0.2.0-alpha.5 at registry `crates-io`
```

After enabling the old protocol, it works as expected:
```
[registries.crates-io]
protocol = "git"
```

imp:
It seems that it is not a freeze, but rather long delay caused by `cargo-release` updating crates.io `git` index (which is falling behind after switch to `sparse` index by default in 1.71). Naturally if `git` index is sufficiently up-to-date the refresh doesn't take long.

Perhaps `crates-index` dependency may be updated to latest version (0.19 -> 2.1) to allow `cargo-release` to utilize `sparse` index instead?

epage:
Closing in favor of #679 (different symptom, same root cause)

Issue #679, opened 2023-05-24
Error when using with CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse


ChriFo:
When enabling the [_new index protocol_](https://blog.rust-lang.org/inside-rust/2023/01/30/cargo-sparse-protocol.html) with e.g. the env variable `CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse` I get the following error when running `cargo release patch`:

```
$ cargo release patch
error: The repo at path /usr/local/cargo/registry/index/github.com-1ecc6299db9ec823 is unusable due to having an invalid HEAD reference: reference 'refs/heads/main' not found; class=Reference (4); code=NotFound (-3)
```

There is no `github.com-*` folder but a directory starting with `index.crates.io`.

cadubentzen:
This also happened to me.
Since the sparse protocol is default as of 1.70.0, this becomes more likely to occur.
Workaround for was setting this in `~/.cargo/config.toml`:
```toml
[registries.crates-io]
protocol = "git"
```

PatOConnor43:
I ran into this using cargo-release 0.24.11. I am also using Rust 1.70.0. I tried adding the `~/.cargo/config.toml` workaround that was suggested above. I wasn't able to successfully `cargo release patch` until I ran `cargo update`. I'm good to go now though.

PR #693, opened 2023-08-22
Unconditionally query crates.io HTTP index for publish status

Jake-Shadle:
This replaces the `crates-index` with `tame-index`, which adds built-in support for sparse registries, resolving a couple of issues, as well as using `gix` instead of `git2` which has improved functionality for authentication, resolving another issue (or at least, I believe it should).

~This unfortunately drastically increases the number of dependencies for this crate, due to gix and reqwest being pulled in, but there are a couple of options to mitigate this.~

~1. Replace the usage of `git2` with `gix` in a follow up PR, which would remove some crates completely and not have 2 implementations of git. There is at least one missing feature in gix that I am aware of that this crate uses (dirty checking of the repository), but since this crate already uses the git cli for some operations like fetch anyways, that would be easy enough to replicate with the cli until gix gets that functionality in the future.~
~2. Just drop support for git registry indices altogether. This may not be possible if using a custom registry that only supports the git sparse protocol, but some alternative registry sources such as cloudsmith already support the sparse protocol, so maybe not that big of an issue. Only using the crates.io sparse registry, even if the user uses an old cargo that doesn't support it, is irrelevant for this crate's usage as it is only checking for the existence of the crate, or the published version.~

Resolves: #218
Resolves: #679
Resolves: #692

### DIFF
--- a/src/ops/cargo.rs
+++ b/src/ops/cargo.rs
@@ -111,7 +111,7 @@ pub fn publish(
 }
 
 pub fn wait_for_publish(
-    index: &mut tame_index::index::ComboIndex,
+    index: &mut crate::steps::index::CratesIoIndex,
     name: &str,
     version: &str,
     timeout: std::time::Duration,
@@ -122,11 +122,6 @@ pub fn wait_for_publish(
         let sleep_time = std::time::Duration::from_secs(1);
         let mut logged = false;
         loop {
-            if let tame_index::index::ComboIndex::Git(gi) = index {
-                if let Err(e) = gi.fetch() {
-                    log::debug!("crate index update failed with {}", e);
-                }
-            }
             if is_published(index, name, version) {
                 break;
             } else if timeout < now.elapsed() {
@@ -147,13 +142,13 @@ pub fn wait_for_publish(
     Ok(())
 }
 
-pub fn is_published(index: &tame_index::index::ComboIndex, name: &str, version: &str) -> bool {
-    match index.krate(name.try_into().expect("crate name is invalid"), true) {
-        Ok(Some(crate_data)) => crate_data
-            .versions
-            .into_iter()
-            .any(|iv| iv.version == version),
-        Ok(None) => false,
+pub fn is_published(
+    index: &mut crate::steps::index::CratesIoIndex,
+    name: &str,
+    version: &str,
+) -> bool {
+    match index.has_krate_version(name, version) {
+        Ok(has_krate_version) => has_krate_version.unwrap_or(false),
         Err(err) => {
             // For both http and git indices, this _might_ be an error that goes away in
             // a future call, but at least printing out something should give the user
--- a/src/ops/cargo.rs
+++ b/src/ops/cargo.rs
@@ -111,7 +111,7 @@ pub fn publish(
 }
 
 pub fn wait_for_publish(
-    index: &mut crates_index::Index,
+    index: &mut tame_index::index::ComboIndex,
     name: &str,
     version: &str,
     timeout: std::time::Duration,
@@ -122,8 +122,10 @@ pub fn wait_for_publish(
         let sleep_time = std::time::Duration::from_secs(1);
         let mut logged = false;
         loop {
-            if let Err(e) = index.update() {
-                log::debug!("crate index update failed with {}", e);
+            if let tame_index::index::ComboIndex::Git(gi) = index {
+                if let Err(e) = gi.fetch() {
+                    log::debug!("crate index update failed with {}", e);
+                }
             }
             if is_published(index, name, version) {
                 break;
@@ -145,12 +147,21 @@ pub fn wait_for_publish(
     Ok(())
 }
 
-pub fn is_published(index: &crates_index::Index, name: &str, version: &str) -> bool {
-    let crate_data = index.crate_(name);
-    crate_data
-        .iter()
-        .flat_map(|c| c.versions().iter())
-        .any(|v| v.version() == version)
+pub fn is_published(index: &tame_index::index::ComboIndex, name: &str, version: &str) -> bool {
+    match index.krate(name.try_into().expect("crate name is invalid"), true) {
+        Ok(Some(crate_data)) => crate_data
+            .versions
+            .into_iter()
+            .any(|iv| iv.version == version),
+        Ok(None) => false,
+        Err(err) => {
+            // For both http and git indices, this _might_ be an error that goes away in
+            // a future call, but at least printing out something should give the user
+            // an indication something is amiss
+            log::warn!("failed to read metadata for {name}: {err:#}");
+            false
+        }
+    }
 }
 
 pub fn set_workspace_version(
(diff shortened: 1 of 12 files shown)


src/steps/index.rs
```diff
@@ -0,0 +1,24 @@
...
+            tame_index::index::RemoteSparseIndex::new(
+                si,
+                tame_index::external::reqwest::blocking::Client::builder()
+                    .http2_prior_knowledge()
+                    .build()
+                    .map_err(tame_index::Error::from)?,
+            )
+            .into()
+        }
+    };
```
epage, line 21: Not too thrilled with the amount of ceremony from this API...

src/ops/cargo.rs
```diff
@@ -145,12 +147,21 @@ pub fn wait_for_publish(
...
 }
 
-pub fn is_published(index: &crates_index::Index, name: &str, version: &str) -> bool {
-    let crate_data = index.crate_(name);
-    crate_data
-        .iter()
-        .flat_map(|c| c.versions().iter())
-        .any(|v| v.version() == version)
+pub fn is_published(index: &tame_index::index::ComboIndex, name: &str, version: &str) -> bool {
+    match index.krate(name.try_into().expect("crate name is invalid"), true) {
```
epage, line 151: Does this call always do a network operation for sparse index?

Inside of cargo, we unconditionally cache lookups in-memory (and on-disk) and it requires an explicit call to force a new lookup.  If the caller has to explicitly decide to do a `krate` call and then `cached_krate`, that seems onerous.

If this doesn't always do a network operation, then this is broken for sparse index as we need to ensure we get the latest published status.
(granted, I should soon just say we don' support those older cargos)

src/ops/cargo.rs
```diff
@@ -122,8 +122,10 @@ pub fn wait_for_publish(
         let sleep_time = std::time::Duration::from_secs(1);
         let mut logged = false;
         loop {
-            if let Err(e) = index.update() {
-                log::debug!("crate index update failed with {}", e);
+            if let tame_index::index::ComboIndex::Git(gi) = index {
+                if let Err(e) = gi.fetch() {
+                    log::debug!("crate index update failed with {}", e);
+                }
```
epage, line 128: What is the locking scheme?

Granted, I'm not correctly handling it for `crates_index` today but its what has stopped me from migrating from v1 to v2 in `cargo-edit` is the lack of clarify on this.

In `cargo-edit`, we've had to implement a retry loop for locking.  I see that tame-index advertises (b)locking support but its unclear to me which locks are involved seeing as tame-index is directly reading and manipulating internal cargo caches (not really thrilled with that btw).  Is it using the same locking policy as cargo?

epage (commented):
Thanks for exploring thos!

src/steps/index.rs
Jake-Shadle: Understandable, but this API was a response to `crates-index` having a much too rigid API that made testing and integration into other crates much more difficult, and some scenarios impossible without replicating pieces of logic that were not public in `crates-index`.

That being said it's trivial to to add helper functions to tame-index to open the crates.io index with zero knobs.

src/ops/cargo.rs
Jake-Shadle: For git it does no network I/O only reading from the cache or git blob if the cache is outdated, for sparse it always does a network call, using the local cache entry to set the e-tag so that the response will only be 200 if the crate metadata has been updated.

src/ops/cargo.rs
Jake-Shadle: So the opening of a git index is locked via gix's mechanism, but tame-index doesn't currently support the global package lock mechanism that cargo uses. I'll open an issue to add that eventually, but I think the easier thing to do for this crate is to simply do one of the options I mentioned in the PR, and just only use the sparse index, and do zero disk operations at all, and remove the gix dependency.

src/steps/index.rs
epage: A lot of libraries have a builder for complex operations and then have a convenience `new` function on the built type.  I would assume something like that would exist here.

src/ops/cargo.rs
Jake-Shadle: Why is that a problem? The current code is doing far more network I/O due to always performing a git fetch, changing it to do a single HTTP request for the crate being released with its etag means drastically reduced and faster checking for the publish status of the crate.

Jake-Shadle:
So I refactored this to still use `tame-index`, but without the `sparse` or `git` features enabled, in favor of doing "manual" HTTP calls to the crates.io index and zero disk I/O, eliminating the concern around cargo's global package lock, and completely sidestepping all of the issues around the git index. This massively reduces the dependencies added by this PR by no longer using gix (though I would be willing to do a follow up PR to replace this crate's usage of git2 with gix since rust > not-rust IMO).

AFAICT there are no tests in this crate around index communication, which makes sense, so I ran a quick test by just releasing one of my own crates with the changes in this PR.

```
   Compiling cargo-fetcher v0.14.6 (/home/jake/code/cargo-fetcher/target/package/cargo-fetcher-0.14.6)
    Finished dev [unoptimized + debuginfo] target(s) in 54.48s
    Packaged 52 files, 415.4KiB (107.8KiB compressed)
   Uploading cargo-fetcher v0.14.6 (/home/jake/code/cargo-fetcher)
    Uploaded cargo-fetcher v0.14.6 to registry `crates-io`
note: Waiting for `cargo-fetcher v0.14.6` to be available at registry `crates-io`.
You may press ctrl-c to skip waiting; the crate should be available shortly.
   Published cargo-fetcher v0.14.6 at registry `crates-io`
     Pushing Pushing 0.14.6, main to origin
```

Obviously you can't tell from the output but the waiting was < 2s, and most importantly, I had blown away `~/.cargo/registry/index/github.com-1ecc6299db9ec823` to ensure that a missing/outdated crates.io git index didn't affect cargo-release, which was my initial reason for opening this PR, seeing that not having a crates.io git index (which will be more and more common due it no longer being the default) caused a massive stall in cargo-release.

Jake-Shadle:
So I had to bump the MSRV, as tame-index requires 1.67.0 due to `gix` also having a 1.67.0 MSRV, if that's an issue I can close this PR.

Jake-Shadle:
Actually I lied, bumped it to 1.67.1 to get https://github.com/rust-lang/rust-clippy/pull/10265 and avoid a bunch of churn that only applies to 1.67.0.

epage:
Finally giving in to going with this approach.  I made some tweaks in #719.

epage:
> So I refactored this to still use `tame-index`, but without the `sparse` or `git` features enabled, in favor of doing "manual" HTTP calls to the crates.io index and zero disk I/O, eliminating the concern around cargo's global package lock, and completely sidestepping all of the issues around the git index.

In doing the upgrades (#722), it drew attention that this wasn't correct and disk IO is done, even in [0.5.1](https://docs.rs/tame-index/0.5.1/tame_index/index/sparse/struct.SparseIndex.html) which this PR is based on.

PR #756, opened 2024-02-25
Use similar rather than difflib to get better diffs

### LINKED ISSUE #755: diff is noisy if you insert new paragraphs into a changelog
In cargo-mutants [Cargo.toml](https://github.com/sourcefrog/cargo-mutants/blob/main/Cargo.toml) I have cargo-release text replacement set up like this:

```toml
[workspace.metadata.release]
pre-release-replacements = [
    { "file" = "NEWS.md", search = "## Unreleased", replace = "$0\n\n## {{version}}", exactly = 1 },
]
```

That is to say it inserts a new version header but leaves an "Unreleased" section at the top.

This works (yay) but I initially thought that it did not, because the displayed diff shows every later paragraph being changed:

```
.... many lines
@@ -565,2 +565 @@
-- Functions that should not be mutated can be marked with `#[mutants::skip]`
-  from the [`mutants`](https://crates.io/crates/mutants) helper crate.
+## 0.0.2
@@ -568 +567,2 @@
-## 0.0.1
+- Functions that should not be mutated can be marked with `#[mutants::skip]`
+  from the [`mutants`](https://crates.io/crates/mutants) helper crate.
@@ -570 +570,3 @@
-First release.
+## 0.0.1
+
+First release.
```

I think this is because the diff code you're using isn't smart about looking for inserted lines. 

It looks like you're using difflib 0.4. I would have guessed it would do better on this, but maybe not. Personally I have used https://crates.io/crates/similar in cargo-mutants and it does seem to get this kind of thing right. I will have a quick look.

sourcefrog:
Gives much better diffs when lines are inserted

Fixes #755

After:

```
   Replacing in NEWS.md
--- NEWS.md
+++ NEWS.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## 24.2.1
+
 - New: `--features`, `--no-default-features` and `--all-features` options are passed through to Cargo.
 
 - Changed: Minimum Rust version (to build cargo-mutants, not to use it) increased to 1.73.
```

If this is OK in principal I can update all the other call sites that use difflib.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,6 +88,7 @@ trycmd = "0.14.21"
 anyhow = "1.0.79"
 concolor-control = { version = "0.0.7", features = ["auto"] }
 git-conventional = "0.12.4"
+similar = "2.4"
 
 [dev-dependencies]
 assert_fs = "1.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,7 +75,6 @@ indexmap = "2.2"
 time = { version = "0.3", features = ["formatting", "macros"] }
 dirs-next = "2.0"
 ignore = "0.4"
-difflib = "0.4"
 itertools = "0.12"
 once_cell = "1.19.0"
 clap = { version = "4.4.18", features = ["derive", "wrap_help"] }
--- a/src/ops/replace.rs
+++ b/src/ops/replace.rs
@@ -1,8 +1,6 @@
 use std::collections::BTreeMap;
 use std::path::Path;
 
-use similar::TextDiff;
-
 use crate::config::Replace;
 use crate::error::CargoResult;
 
@@ -121,14 +119,12 @@ pub fn do_file_replacements(
         if data != replaced {
             if dry_run {
                 if noisy {
-                    let diff = TextDiff::from_lines(&data, &replaced);
-                    let display_path = path.display().to_string();
                     let _ = crate::ops::shell::status(
                         "Replacing",
                         format!(
                             "in {}\n{}",
                             path.display(),
-                            diff.unified_diff().header(&display_path, &display_path)
+                            crate::diff::unified_diff(&data, &replaced, &path, "replaced")
                         ),
                     );
                 } else {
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -88,6 +88,7 @@ trycmd = "0.14.21"
 anyhow = "1.0.79"
 concolor-control = { version = "0.0.7", features = ["auto"] }
 git-conventional = "0.12.4"
+similar = "2.4"
```
epage, line 91: If we go forward with this, could you completely replace difflib and remove the dependency on it?

epage: Finished looking over it and it seems it would be reasonable to move forward with for now.

sourcefrog: > If we go forward with this, could you completely replace difflib and remove the dependency on it?

Yes, I will, I just wanted to check it was basically ok first.

sourcefrog: Done

src/ops/replace.rs
```diff
@@ -118,25 +120,15 @@ pub fn do_file_replacements(
...
                 if noisy {
+                    let diff = TextDiff::from_lines(&data, &replaced);
+                    let display_path = path.display().to_string();
                     let _ = crate::ops::shell::status(
                         "Replacing",
                         format!(
                             "in {}\n{}",
                             path.display(),
-                            itertools::join(diff.into_iter(), "")
+                            diff.unified_diff().header(&display_path, &display_path)
```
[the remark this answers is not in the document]
sourcefrog: Done

sourcefrog:
Thanks, please take another look.

Incidentally did you mean for there to be two `shell.rs` with almost identical contents? I didn't need to change them for this but I just noticed it and wondered if it was due to an accident with git or something.

sourcefrog:
The `cargo_deny` failure (https://github.com/crate-ci/cargo-release/pull/756/checks) seems unrelated to this change

epage:
Fixed `shell.rs` in a follow up.

PR #787, opened 2024-05-09
Enable selecting system certificate stores.

### LINKED ISSUE #764: "Invalid peer certificate" when run on network which substitutes certs
This issue appears to have arisen between `v0.20.5` and `v0.25.6` (it worked for me on the former version, and now doesn't on the latter).

Basically, I'm running on a network where certificates for sites are substituted with a chain managed by the network operator. I have the root for this chain installed in my system certificate store.

I can confirm with the `-vv` flag that `rustls` under the hood _is_ picking up my network's substitute chain, which it then fails to verify.

It looks like `cargo-release` uses `webpki_roots` through `reqwest`, which only considers the Mozilla root certificate list, and _does not_ incorporate what certs are trusted in the system store.

I believe the solution would be to modify how `reqwest` is used to incorporate `rustls_native_certs` instead.

alilleybrinker:
This commit adds the ability to select what certificate root store to use when making HTTPS connections. Previously, `cargo-release` was configured to use the default for `tame-index`, which is the "webpki" root of trust maintained by Mozilla. This is a good and reasonable set of certificates, _except_ for users on networks which substitute certificates.

Substituting certificates is something enterprise networks will frequently do so they can man-in-the-middle HTTPS connections made on their network and thus maintain visibility into the network activities of their employees. In this setup, the users' devices will generally be running enterprise-managed software which replaces certificates used by public websites with ones provided by the network software the enterprise uses, with the root certificates for these substituted chains being placed in the users' local system certificate store. In that case, with only the "webpki" certificate store loaded for `cargo-release`, the substituted certificates will fail to validate, and publication of new versions (indeed, even checking publication status of the crate attempting to be published) will fail with an HTTPS error about an untrusted certificate.

The solution chosen here was to add a configuration element, and a CLI flag under the "publish" section, which lets the user pick between "webpki" (Mozilla) or "native" (local system) certificate trust stores. The portion of the code in `src/ops/index.rs` which handles connecting to the registry index has been updated to configure which certificate store to use based on the user's selection.

The one final wrinkle is that we get `reqwest`, the dependency which actually handles HTTPS connections, through the `tame-index` crate, which re-exports it. To enable the APIs in `reqwest` for configuring what TLS certs to pick up, we have to enable the "native-certs" feature on `tame-index`, while leaving the default features on. This is something `tame-index` normally recommends _against_, because it assumes you want to exclusively activate one of them at compile time. In our case, we need the selection to happen at runtime, so we need both to be compiled in.

Fixes #764

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -27,25 +27,27 @@ Arguments:
                    values: major, minor, patch, release, rc, beta, alpha]
 
 Options:
-      --manifest-path <PATH>        Path to Cargo.toml
-  -p, --package <SPEC>              Package to process (see `cargo help pkgid`)
-      --workspace                   Process all packages in the workspace
-      --exclude <SPEC>              Exclude packages from being processed
-      --unpublished                 Process all packages whose current version is unpublished
-  -m, --metadata <METADATA>         Semver metadata
-  -x, --execute                     Actually perform a release. Dry-run mode is the default
-      --no-confirm                  Skip release confirmation and version preview
-      --prev-tag-name <NAME>        The name of tag for the previous release
-  -c, --config <PATH>               Custom config file
-      --isolated                    Ignore implicit configuration files
-      --sign                        Sign both git commit and tag
-      --dependent-version <ACTION>  Specify how workspace dependencies on this crate should be
-                                    handed [possible values: upgrade, fix]
-      --allow-branch <GLOB[,...]>   Comma-separated globs of branch names a release can happen from
-  -q, --quiet...                    Pass many times for less log output
-  -v, --verbose...                  Pass many times for more log output
-  -h, --help                        Print help (see more with '--help')
-  -V, --version                     Print version
+      --manifest-path <PATH>         Path to Cargo.toml
+  -p, --package <SPEC>               Package to process (see `cargo help pkgid`)
+      --workspace                    Process all packages in the workspace
+      --exclude <SPEC>               Exclude packages from being processed
+      --unpublished                  Process all packages whose current version is unpublished
+  -m, --metadata <METADATA>          Semver metadata
+  -x, --execute                      Actually perform a release. Dry-run mode is the default
+      --no-confirm                   Skip release confirmation and version preview
+      --prev-tag-name <NAME>         The name of tag for the previous release
+  -c, --config <PATH>                Custom config file
+      --isolated                     Ignore implicit configuration files
+      --sign                         Sign both git commit and tag
+      --dependent-version <ACTION>   Specify how workspace dependencies on this crate should be
+                                     handed [possible values: upgrade, fix]
+      --allow-branch <GLOB[,...]>    Comma-separated globs of branch names a release can happen from
+      --certs-source <CERTS_SOURCE>  Indicate what certificate store to use for web requests
+                                     [possible values: webpki, native]
+  -q, --quiet...                     Pass many times for less log output
+  -v, --verbose...                   Pass many times for more log output
+  -h, --help                         Print help (see more with '--help')
+  -V, --version                      Print version
 
 Commit:
       --sign-commit  Sign git commit
@@ -140,7 +142,7 @@ Workspace configuration is read from the following (in precedence order)
 | `metadata`     | \-              | `optional`, `required`, `ignore`, `persistent` | `optional` | Policy for presence of absence of `--metadata` flag when changing the version |
 | `rate-limit.new-packages` | \-   | integer                     | `5`           | `optional` | Rate limit for publishing new packages |
 | `rate-limit.existing-packages` | \- | integer                  | `30`          | `optional` | Rate limit for publishing existing packages |
-
+| `certs-source` | \-              | `webpki`, `native`                             | `webpki`   | Policy for using Mozilla's standard certificate root of trust (`webpki`) or using the system certificate root of trust (`native`) |
 
 Note: fields are from the package-configuration unless otherwise specified.
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -38,6 +38,7 @@ pub struct Config {
     pub metadata: Option<MetadataPolicy>,
     pub target: Option<String>,
     pub rate_limit: RateLimit,
+    pub certs_source: Option<CertsSource>,
 }
 
 impl Config {
@@ -87,6 +88,7 @@ impl Config {
             metadata: Some(empty.metadata()),
             target: None,
             rate_limit: RateLimit::from_defaults(),
+            certs_source: Some(empty.certs_source()),
         }
     }
 
@@ -166,8 +168,10 @@ impl Config {
         if let Some(target) = source.target.as_deref() {
             self.target = Some(target.to_owned());
         }
-
         self.rate_limit.update(&source.rate_limit);
+        if let Some(certs) = source.certs_source {
+            self.certs_source = Some(certs);
+        }
     }
 
     pub fn allow_branch(&self) -> impl Iterator<Item = &str> {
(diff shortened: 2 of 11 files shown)


src/ops/index.rs
```diff
@@ -83,13 +91,24 @@ pub struct RemoteIndex {
...
+        // Use a full 'match' here to prepare for the possible future of needing more
+        // complex cert-selection logic, for example wanting to include _both_
+        // certificate stores.
+        let (webpki_certs, native_certs) = match cert_selection {
+            CertSelection::Webpki => (true, false),
+            CertSelection::Native => (false, true),
+        };
+
         let client = tame_index::external::reqwest::blocking::ClientBuilder::new()
-            .http2_prior_knowledge()
```
alilleybrinker, line 91: With native certs on, this line caused errors. This line turns off ALPN, so there's no ability to negotiate between HTTP versions. I am not sure _why_ this was causing things to fail, but I kept getting errors with it on. The difference for removing this line is that ALPN is turned _on_, so HTTP version negotiation can occur.

src/ops/index.rs
```diff
@@ -83,13 +91,24 @@ pub struct RemoteIndex {
...
             tame_index::IndexUrl::CratesIoSparse,
         ))?;
+
+        // Use a full 'match' here to prepare for the possible future of needing more
+        // complex cert-selection logic, for example wanting to include _both_
+        // certificate stores.
+        let (webpki_certs, native_certs) = match cert_selection {
+            CertSelection::Webpki => (true, false),
+            CertSelection::Native => (false, true),
+        };
```
alilleybrinker, line 105: Tried to reflect it in the comments here: this match is basically defensive coding against the possibility that more `CertSelection` variants may be added in the future. If they are, we want to make sure they're reflected in different certificate selection logic here.

docs/reference.md
```diff
@@ -112,35 +112,35 @@ Workspace configuration is read from the following (in precedence order)
...
-| `enable-all-features` | `--all-features` | bool                | `false`       | Signal to `cargo publish`, that all features should be used (requires rust 1.33+) |
-| `target`       | \-              | string                      | \-            | Target triple to use for the verification build |
-| `dependent-version` | \-         | `upgrade`, `fix`, `error`, `warn`, `ignore` | `upgrade`      | Policy for upgrading path dependency versions within the workspace |
-| `metadata`     | \-              | `optional`, `required`, `ignore`, `persistent` | `optional` | Policy for presence of absence of `--metadata` flag when changing the version |
-
+| Field                        | Argument          | Format                                         | Defaults                                              | Description                                                                                                                                                                                                                              |
```
epage, line 115: If you are reformatting the markdown table, please isolate that to its own commit

docs/reference.md
```diff
@@ -57,6 +57,8 @@ Publish:
       --features <FEATURES>  Provide a set of features that need to be enabled
       --all-features         Enable all features via `all-features`. Overrides `features`
       --target <TRIPLE>      Build for the target triple
+      --use-certs <CERTS>    Select what certificates to use (default: 'webpki') [possible values:
```
epage, line 60: Is there precedence with the `user-certs` name in other applications?  For me, it feels a little weird.  I lean towards `certs` or `certs-source`.

Thoughts?

src/ops/index.rs
```diff
@@ -83,13 +91,24 @@ pub struct RemoteIndex {
...
     #[inline]
-    pub fn open() -> Result<Self, crate::error::CliError> {
+    pub fn open(cert_selection: CertSelection) -> Result<Self, crate::error::CliError> {
         let index = tame_index::SparseIndex::new(tame_index::IndexLocation::new(
             tame_index::IndexUrl::CratesIoSparse,
         ))?;
+
+        // Use a full 'match' here to prepare for the possible future of needing more
+        // complex cert-selection logic, for example wanting to include _both_
+        // certificate stores.
```
epage, line 101: imo this comment could be removed

src/ops/index.rs
```diff
@@ -83,13 +91,24 @@ pub struct RemoteIndex {
...
+        let (webpki_certs, native_certs) = match cert_selection {
+            CertSelection::Webpki => (true, false),
+            CertSelection::Native => (false, true),
+        };
+
         let client = tame_index::external::reqwest::blocking::ClientBuilder::new()
-            .http2_prior_knowledge()
+            .tls_built_in_webpki_certs(webpki_certs)
```
epage, line 108: What if we did
```rust
let client = tame_index::external::reqwest::blocking::ClientBuilder::new();
let client = match cert_selection{
   ...
};
let client = client.build()?;
```

docs/reference.md
alilleybrinker: `certs-source` seems better to me! I'll make the adjustment.

alilleybrinker:
Ended up opting to rebase down to one commit, and also fixed the table reformatting. If you'd still like that in its own commit separately, let me know.

src/config.rs
```diff
@@ -85,6 +86,7 @@ impl Config {
             dependent_version: Some(empty.dependent_version()),
             metadata: Some(empty.metadata()),
             target: None,
+            certs_source: None,
```
epage, line 89: This should cal `Some(empty.certs_source())`

src/config.rs
```diff
@@ -630,6 +651,10 @@ pub struct PublishArgs {
     /// Build for the target triple
     #[arg(long, value_name = "TRIPLE")]
     target: Option<String>,
+
+    /// Select what certificates to use (default: 'webpki')
+    #[arg(long, value_name = "CERTS")]
+    certs_source: Option<CertsSource>,
```
epage, line 657: This feels weird.  We are limiting this flag to just `PublishArgs` but most commands need to know the certs source so users will have to go through config and can't use the command-line.

We should either remove the flag or have it on all commands.

docs/reference.md
```diff
@@ -27,38 +27,38 @@ Arguments:
...
-      --isolated                    Ignore implicit configuration files
-      --sign                        Sign both git commit and tag
-      --dependent-version <ACTION>  Specify how workspace dependencies on this crate should be
-                                    handed [possible values: upgrade, fix]
-      --allow-branch <GLOB[,...]>   Comma-separated globs of branch names a release can happen from
-  -q, --quiet...                    Pass many times for less log output
-  -v, --verbose...                  Pass many times for more log output
-  -h, --help                        Print help (see more with '--help')
-  -V, --version                     Print version
+      --manifest-path <PATH>         Path to Cargo.toml
```
epage, line 30: Can you squash this into the appropriate commit?

alilleybrinker:
Rebased on upstream. Doing local tests now and may have additional fixes to keep things running. Prior [security audit](https://github.com/crate-ci/cargo-release/actions/runs/9568133724/job/26526737746) failed so I'd like to address that.

docs/reference.md
alilleybrinker: Would squashing all commits down to one be sufficient?

alilleybrinker:
Regarding the `cargo-audit` CI failure: it looks `cargo-release` itself and `cargo-test-support` are trying to link in different versions of `libgit2`. I can't reproduce this locally, and it's not tied to the changes here which don't modify any of the relevant dependencies which would induce this issue. I'd argue it's not something to try to resolve here, but ought to be resolved in `crate-ci:master` and then I can rebase to get the fix in this branch.

docs/reference.md
epage: If it makes sense within the commits, yes

epage:
The audit is non-blocking

docs/reference.md
alilleybrinker: Yeah, this change is to fix a test breakage caused by the `--certs-source` flag changing how `--help` formats the output spacing overall. I've squashed it all down to one.

Issue #790, opened 2024-05-23
failing to release with more than 30 crates


rnbguy:
The concerning code is below:

https://github.com/crate-ci/cargo-release/blob/4b09269f320d2186b58a233306ecd6c1b848b746/src/steps/mod.rs#L251-L261

Not sure, how this hard-coded number, 30, is derived in #545. AFAIK, it is much higher than 30 (for publishing existing crates). Can we at least make it flexible - maybe via an environment variable?

epage:
> Not sure, how this hard-coded number, 30, is derived in https://github.com/crate-ci/cargo-release/pull/545. AFAIK, it is much higher than 30 (for publishing existing crates). Can we at least make it flexible - maybe via an environment variable?

The limits came from an email someone got from the crates.io team

> The rate limit for creating new crates is 1 crate every 10 minutes, with a burst of 5 crates.
>
> The rate limit for new versions of existing crates is 1 per minute, with a burst of 30 crates, so when releasing new versions of these crates, you shouldn't hit the limit.
>
> It's not particularly secret, we just don't publish it other than in the code because it's subject to change. The responses from the rate limited requests on when to try again contain the most accurate information.


See https://github.com/crate-ci/cargo-release/issues/483#issuecomment-1173038971

They could have adjusted the base limits since 2022.  They also adjust the limits for known special cases.

rnbguy:
Totally okay with putting this in cargo-release config :+1:

```toml
[workspace.metadata.release]
ratelimit = { existing = 50 }
```

I would be up for a PR. Just let me know :slightly_smiling_face:

epage:
Seems reasonable.  Go for it :)

PR #792, opened 2024-05-29
feat: support project level `rate_limit`

rnbguy:
Closes: #790

With this PR, we should be able to include this in a workspace `Cargo.toml`:

```toml
[workspace.metadata.release]
rate-limit = { existing-packages = 50 }
```

This would allow a maximum of 50 crates to be published together. This is currently hard-coded to be 30.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -138,8 +138,8 @@ Workspace configuration is read from the following (in precedence order)
 | `target`       | \-              | string                      | \-            | Target triple to use for the verification build |
 | `dependent-version` | \-         | `upgrade`, `fix`, `error`, `warn`, `ignore` | `upgrade`      | Policy for upgrading path dependency versions within the workspace |
 | `metadata`     | \-              | `optional`, `required`, `ignore`, `persistent` | `optional` | Policy for presence of absence of `--metadata` flag when changing the version |
-| `rate-limit.new-packages` | \-   | integer                     | `5`           | `optional` | Rate limit for publishing new packages |
-| `rate-limit.existing-packages` | \- | integer                  | `30`          | `optional` | Rate limit for publishing existing packages |
+| `rate-limit.new-packages` | \-   | positive integer            | `5`           | `optional` | Rate limit for publishing new packages |
+| `rate-limit.existing-packages` | \- | positive integer         | `30`          | `optional` | Rate limit for publishing existing packages |
 
 
 Note: fields are from the package-configuration unless otherwise specified.
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -138,7 +138,8 @@ Workspace configuration is read from the following (in precedence order)
 | `target`       | \-              | string                      | \-            | Target triple to use for the verification build |
 | `dependent-version` | \-         | `upgrade`, `fix`, `error`, `warn`, `ignore` | `upgrade`      | Policy for upgrading path dependency versions within the workspace |
 | `metadata`     | \-              | `optional`, `required`, `ignore`, `persistent` | `optional` | Policy for presence of absence of `--metadata` flag when changing the version |
-| `rate-limit`   | \-              | maps of strings to integers | `{ new-packages = 5, existing-packages = 30 }` | `optional` | Rate limits for the release process |
+| `rate-limit.new-packages` | \-   | integer                     | `5`           | `optional` | Rate limit for publishing new packages |
+| `rate-limit.existing-packages` | \- | integer                  | `30`          | `optional` | Rate limit for publishing existing packages |
 
 
 Note: fields are from the package-configuration unless otherwise specified.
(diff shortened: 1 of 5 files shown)


epage:
Could you clean up your commit to being atomic?

src/steps/mod.rs
```diff
@@ -236,26 +237,40 @@ pub fn verify_rate_limit(
         }
     }
 
-    if 5 < new {
+    let rate_limit_config = ws_config.rate_limit();
+
+    if rate_limit_config.new.unwrap_or(5) < new {
```
epage, line 242: Normally we handle defaults by
- Having a `fn new(&self) -> usize`
- Having `RateLimit::from_defaults` call `RateLimit::new`

So here you'd just call `RateLimit::new`

src/steps/mod.rs
```diff
@@ -236,26 +237,40 @@ pub fn verify_rate_limit(
...
-                "attempting to publish {} new crates which is above the crates.io rate limit",
-                new
+                "attempting to publish {} new crates which is above the {} rate limit: {}",
+                new,
+                if rate_limit_config.new.is_some() {
+                    "project"
+                } else {
+                    "crates.io"
+                },
+                rate_limit_config.new.unwrap_or(5)
```
epage, line 255: Let's just remove the source

src/config.rs
```diff
@@ -452,6 +462,12 @@ struct CargoMetadata {
     release: Option<Config>,
 }
 
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct RateLimit {
+    pub new: Option<usize>,
+    pub existing: Option<usize>,
```
epage, line 468: Especially seeing the function names, I'm thinking we should be more specific

Some ideas
- `num_new` / `num_existing`
- `new_packages` / `existing_packages`

Thoughts?

src/config.rs
```diff
@@ -452,6 +462,12 @@ struct CargoMetadata {
     release: Option<Config>,
 }
 
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct RateLimit {
```
epage, line 466: Please add
```rust
#[serde(rename_all = "kebab-case")]
```

epage, line 466: `docs/reference.md` needs to be updated to reference this

rnbguy: f38d5ae

src/config.rs
rnbguy: f38d5ae

src/steps/mod.rs
rnbguy: bf6979a

src/config.rs
rnbguy: 0560aac

src/steps/mod.rs
rnbguy: Doe this looks good?

c236baf

epage: This put the functions on `Config`.  Is there a reason you didn't put them on `RateLimit` like I mentioned?

src/config.rs
epage: The reference should have a line per field, so they'd be `rate-limit.new-packages` and `rate-limit.existing-packages`.  Documenting them together makes it a lot harder for the reader to understand what they are supposed to do.

rnbguy: Makes sense. 8f998ff

src/steps/mod.rs
rnbguy: I think, I didn't understand you. :sweat_smile: I refactored the code once again. Does it look better?

b13b6ab

rnbguy: This doesn't work correctly. I will take a look next week.

docs/reference.md
```diff
@@ -138,6 +138,8 @@ Workspace configuration is read from the following (in precedence order)
 | `target`       | \-              | string                      | \-            | Target triple to use for the verification build |
 | `dependent-version` | \-         | `upgrade`, `fix`, `error`, `warn`, `ignore` | `upgrade`      | Policy for upgrading path dependency versions within the workspace |
 | `metadata`     | \-              | `optional`, `required`, `ignore`, `persistent` | `optional` | Policy for presence of absence of `--metadata` flag when changing the version |
+| `rate-limit.new-packages` | \-   | positive integer            | `5`           | `optional` | Rate limit for publishing new packages |
+| `rate-limit.existing-packages` | \- | positive integer         | `30`          | `optional` | Rate limit for publishing existing packages |
```
epage, line 142: In documentation, we have to ensure we give the user the information they need without overwhelming them, making it harder to find.  I feel like `integer` is sufficient here.

src/config.rs
```diff
@@ -452,6 +460,43 @@ struct CargoMetadata {
...
+}
+
+impl RateLimit {
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn from_defaults() -> Self {
+        Self::new()
+    }
```
epage, line 488: This package doesn't have too many useful examples of the pattern I'm talking about.

See https://github.com/crate-ci/typos/blob/master/crates/typos-cli/src/config.rs
- All fields should be `Option` (`0` is not an appropriate sentinel in Rust for "not set")
- `Default` should be derived
- `RateLimit::from_defaults` is default configuration which is different than Default initialization

docs/reference.md
rnbguy: fair enough :+1:

760b3e0

src/config.rs
rnbguy: Thanks for the reference and explanation. It helped.

ee1d90c

epage:
Thanks!

PR #811, opened 2024-08-22
Readme: Fix heading

Turbo87:
I assume this was meant to be a heading instead of a regular paragraph :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # cargo release
 
-Features
+## Features
+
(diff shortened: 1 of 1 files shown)


epage:
I had intended it to be a regular text intro to the bulleted list.  Is there a reason you prefer a header for it?

Turbo87:
idk, it's just what I would've expected with "Install" and "Usage" also being headings

PR #820, opened 2024-09-23
feat: Out more error info with cmd run failed

richerfu:
There are two reason for this pr:
1. `std::process.Command` recommended us to use canonicalize instead of current_dir [doc](https://doc.rust-lang.org/std/process/struct.Command.html#method.current_dir)
2. Convert to absolute path make us easier to locate problems.
    Before:
    image
    After:
    image

### DIFF
--- a/src/ops/cmd.rs
+++ b/src/ops/cmd.rs
@@ -38,12 +38,17 @@ fn do_call(
         }
     }
 
+    let ctx_dir = match path {
+        Some(p) => format!(" within {}", p.display()),
+        None => String::new(),
+    };
+
     let mut child = cmd
         .spawn()
-        .map_err(|e| anyhow::format_err!("failed to launch `{cmd_name}`: {e}"))?;
+        .map_err(|e| anyhow::format_err!("failed to launch `{cmd_name}`{ctx_dir}: {e}"))?;
     let result = child
         .wait()
-        .map_err(|e| anyhow::format_err!("failed to launch `{cmd_name}`: {e}"))?;
+        .map_err(|e| anyhow::format_err!("failed to launch `{cmd_name}`{ctx_dir}: {e}"))?;
 
     Ok(result.success())
 }


src/ops/cmd.rs
```diff
@@ -20,13 +20,19 @@ fn do_call(
...
+        |p| {
+            dunce::canonicalize(p.join(origin_cmd_name))
+                .map(|path| path.to_string_lossy().into_owned())
+                .unwrap_or_else(|_| p.join(origin_cmd_name).to_string_lossy().into_owned())
+        },
+    );
 
-    if let Some(p) = path {
-        cmd.current_dir(p);
-    }
```
epage, line 29: This is a breaking change as the script might rely on things based on the relative path

src/ops/cmd.rs
```diff
@@ -20,13 +20,19 @@ fn do_call(
...
     let mut iter = command.iter();
-    let cmd_name = iter.next().unwrap();
+    let origin_cmd_name: &String = iter.next().unwrap();
 
-    let mut cmd = Command::new(cmd_name);
+    // convert to absolute path if path is existed as first and avoid to use Command::current_dir
+    let cmd_name = path.map_or_else(
+        || origin_cmd_name.clone(),
+        |p| {
+            dunce::canonicalize(p.join(origin_cmd_name))
```
epage, line 29: > Convert to absolute path make us easier to locate problems.

imo showing absolute paths in error messages also makes things harder for users. I wonder if the better option is to improve the error message.

richerfu: Yep, so how about to try it below and revert `cmd_name` logic:
```rust
    let ctx_dir = path.map_or(String::new(), |p| format!(" within {:?}", p));
    
    let mut child = cmd
        .spawn()
        .map_err(|e| anyhow::format_err!("failed to launch `{cmd_name}`{ctx_dir}: {e}"))?;
    let result = child
        .wait()
        .map_err(|e| anyhow::format_err!("failed to launch `{cmd_name}`{ctx_dir}: {e}"))?;
```

epage: That general idea, sure.  Note you should use `p.display()` instead of `{:?}`.

epage:
Could you clean up the commits for how you would want them presented for review and merging?

richerfu:
> Could you clean up the commits for how you would want them presented for review and merging?

Cleaned and all looks good for me. Please go to next step, thanks

richerfu:
> Could you clean up the commits for how you would want them presented for review and merging?

Cleaned and all looks good for me. Please go to the next step thanks.

epage:
Thanks!

PR #840, opened 2024-11-08
Update README.md

tusharmath:
Copy paste should just simply work instead of pasting and then removing `$` everytime after pasting.


epage:
This is needed for the `console` code block to be formatted correctly

PR #848, opened 2024-12-21
Update README.md

Ghaniyyat05:
Added short description for the project


epage:
Thanks for pointing out this was missing and putting the effort in writing this!  In the end, I went my own route for how this was written to match my style.

PR #857, opened 2025-01-20
chore(deps): Drop anstyle crate for clap_cargo::style

### LINKED ISSUE #856: Consider Using `clap_cargo::style` Instead of `anstyle`
I noticed the project is using `anstyle` for styling. Since `clap_cargo::style` is already part of the `clap` ecosystem and offers styling, it might be worth considering switching to it. This could simplify things by keeping the dependencies more cohesive with `clap`.

If there’s a specific reason for using `anstyle` that I’ve missed, feel free to close this issue!

xosnrdev:
This closes #856


src/ops/shell.rs
```diff
@@ -1,10 +1,9 @@
-use std::io::{stdin, stdout, Write};
-
-use anstyle::Style;
 use anyhow::Context as _;
+use clap::builder::styling::Style;
+use clap_cargo::style::{ERROR, HEADER, NOTE, WARN};
+use std::io::{stdin, stdout, Write};
```
epage, line 5: This changed from our existing style which is
1. `std::`
2. external
3. `crate::`

xosnrdev: `cargo fmt` did the reordering.

xosnrdev: Resolved.

epage: Not by default.  Either a blank line needs to be removed or a user config is present.  For the latter, I'd recommend avoiding that as it can cause problems with contributing to different projects.

xosnrdev: Yea i figured.

PR #898, opened 2025-09-06
docs: Use mdsh to keep reference doc up-to-date

lafrenierejm:
This is partially a follow-up to #840; the `cargo install` block is changed to a raw shell command for easier copy-paste.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,3 +17,8 @@ repos:
     rev: v1.1.7
     hooks:
     - id: committed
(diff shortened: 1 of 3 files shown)


epage:
We generally ask for problems and potential solutions to be discussed in Issues as it is the more likely place to find them and doing it in PRs splits up the conversation.  The fact that we have #840 and this is an example of that problem. Replacing current processes (`trycmd`) with new ones (`mdsh`) is another example  of something that should likely be socialized.

PR #954, opened 2026-04-16
fix: avoid panic when hook is empty array

LuoboTian:
Fixes #945

## Problem
When configuring `pre-release-hook = []` (empty array), the tool would panic with `unwrap` on a `None` value.

## Solution
Add an early return when the command vector is empty, which:
- Prevents the panic
- Treats empty hooks as "no operation" (standard CLI behavior)
- Is a minimal, safe change with no breaking changes

## Testing
- Verified with `pre-release-hook = []`
- No panic, runs successfully
- Dry-run and normal modes both work correctly

### DIFF
--- a/src/ops/cmd.rs
+++ b/src/ops/cmd.rs
@@ -12,6 +12,11 @@ fn do_call(
     dry_run: bool,
 ) -> CargoResult<bool> {
     let command: Vec<_> = command.into_iter().map(|s| s.into()).collect();
+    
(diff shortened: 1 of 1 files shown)


epage:
As this does not follow our [contrib guide](https://github.com/crate-ci/cargo-release/blob/master/CONTRIBUTING.md#pull-requests) and looks to be a low effort LLM PR, I'm going to close this.

PR #967, opened 2026-05-13
fix: use workspace root manifest for cargo publish

shirok1:
Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/steps/publish.rs
+++ b/src/steps/publish.rs
@@ -152,7 +152,7 @@ impl PublishStep {
         super::confirm("Publish", &selected_pkgs, self.no_confirm, dry_run)?;
 
         // STEP 3: cargo publish
-        publish(&selected_pkgs, dry_run)?;
+        publish(&selected_pkgs, &ws_meta, dry_run)?;
 
         super::finish(failed, dry_run)
     }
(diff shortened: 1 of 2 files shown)


epage:
This says the "what" but not the "why".

Note: our contrib guide asks for the "why" to be discussed in issues before posting PRs.

shirok1:
Currently the manifest path passed to `cargo publish` is unpredictably selected by `pkgs.first()`, which is:

1. not align with rest of `cargo` op calls, which use workspace discovered by `cargo metadata` wherever possible
2. causing unpredictable release failure in my scenario, more on that below

I'm working on a project with 2 workspaces (one for open sourcing and one for internal). I need to publish crates IN the open source workspace WITH the internal workspace. The directory structure looks like this:

```
❯ eza --tree --level 2
.
├── Cargo.toml # [workspace.dependencies] crate_a = { path = "oss/crate_a", registry = "xxx", version = "0.2" }
├── crates
│   ├── xxx
│   ├── xxx
├── oss
│   ├── Cargo.toml # [workspace.dependencies] crate_a = { path = "crate_a", registry = "xxx", version = "0.2" }
│   └── crate_a
│       ├── Cargo.toml
│       └── src
└── crate_b
    ├── Cargo.toml # crate_a = { workspace = true }
    └── src
```

If `pkgs.first()` selected `crate_b` then top-level workspace will be used and `crate_b` can be uploaded, but if `crate_a` is selected instead `cargo publish` will complain the missing of `crate_b` not to mention other deps defined with `workspace = true` will point to wrong spec

epage:
Again, these kinds of details should be engaged with on an issue, not a PR

shirok1:
Fixes #1014

PR #968, opened 2026-05-13
feat: allow {{version}} in consolidated commit message

shirok1:
- **feat: resolve {{version}} in consolidated commit message from common package version**
- **fix: skip crates.io index lookup for non-published packages**
- **test: add integration test for {{version}} in consolidated commit message**


epage:
This has been rejected in many issues already. The path forward is to find one with the reason given and to suggest a design or provide reasoning that addresses the concern.

shirok1:
I suggest you mean discussion in #246, and your response was `shared-version = true`. However in our monorepo we put WASM framework and native VM extension in the same workspace, and they have different version managing methods and timelines (WASM framework on private registry and native things locked to commits of repo when building our apps). I wrote scripts so I can publish one of them (say sub-set of crates) with `cargo release` separately using a long list of "--package xxx", so each group have its own semvers. I don't think `shared-version = true` is the right semantic here, cuz it really isn't something "workspace-level".

epage:
PRs are not the place to discuss use cases, design ideas, etc. To encourage that, such comments will be ignored.

PR #970, opened 2026-05-13
feat(cli): add --pre-release-commit-message arg to override commit message template

shirok1:


epage:
As the contrib guide says, we should start with an issue that establishes the problem being solved and where we explore solutions and decide on one.

shirok1:
Our CI require ticket ID on e.g. Jira in commit messages, so instead of `chore: Release` we need `chore-11111: Release`. I wrote a script to extract this info from git branch name. Since `reference.md` got so many other config "in precedence order" I think its actually counterintuitive to not have this arg.

epage:
I stick with the github native route of having the issue in a footer. I don't enforce it because there are (rare) times where an Issue isn't needed.

shirok1:
Fixes #971

Issue #983, opened 2026-06-03
False positive warning on "Unrendered {{version}} ..."


tisonkun:
```
cargo release -p logforth -x
Release logforth 0.30.0? [y/N] 
y
[WARN ] Unrendered {{version}} present in template "chore: Release {{crate_name}} version {{version}}"
[WARN ] Unrendered {{crate_name}} present in template "chore: Release {{crate_name}} version {{version}}"
  Publishing logforth
    Updating crates.io index
   Packaging logforth v0.30.0 (/path/to/logforth/logforth)
    Updating crates.io index
    Packaged 9 files, 85.6KiB (21.0KiB compressed)
   Verifying logforth v0.30.0 (/path/to/logforth/logforth)
   Compiling logforth v0.30.0 (/path/to/logforth/target/package/logforth-0.30.0)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.85s
   Uploading logforth v0.30.0 (/path/to/logforth/logforth)
    Uploaded logforth v0.30.0 to registry `crates-io`
note: waiting for logforth v0.30.0 to be available at registry `crates-io`
help: you may press ctrl-c to skip waiting; the crate should be available shortly
   Published logforth v0.30.0 at registry `crates-io`
     Pushing Pushing logforth-v0.30.0 to origin
Enumerating objects: 1, done.
Counting objects: 100% (1/1), done.
Writing objects: 100% (1/1), 852 bytes | 852.00 KiB/s, done.
Total 1 (delta 0), reused 0 (delta 0), pack-reused 0 (from 0)
To https://github.com/fast/logforth.git
 * [new tag]         logforth-v0.30.0 -> logforth-v0.30.0
```

However, the variable has been rendered at: https://github.com/fast/logforth/releases/tag/logforth-v0.30.0

cargo-release's metadata is defined at: https://github.com/fast/logforth/blob/4542ccf64fa2febd761352318c4cf18e214f028c/Cargo.toml#L87-L91

epage:
From your pointing at the tag, it seems like you are thinking the message is warning about the tag. I think it instead is warning about `pre-release-commit-message` but you had no content in the commit, so it did nothing, so you don't see the unrendered template variables.

On one hand, this is a false positive because the template is unused. We'd need to add more change tracking to decide whether to bother rendering the template.

On the other hand, by us always reporting this, it makes it easier to catch the problem. It would help if we said what template was being rendered.

tisonkun:
[...]

PR #984, opened 2026-06-03
fix: skip commit if Git status is clean

tisonkun:
This refers to #983.


epage:
Not completely sold on this change as noted in the issue

> On the other hand, by us always reporting this, it makes it easier to catch the problem.

Discussing that further should happen on the issue. Moving to a draft for now (as our contrib docs ask not even to start on a PR until there is buy in)

PR #996, opened 2026-07-09
fix: preserve TOML value decoration

### LINKED ISSUE #671: Adds a space
The most minor issue ever, but adding since hopefully helpful on net.

Bumping releases adds a space after the version number:

```diff
diff --git a/prql-compiler/prqlc/Cargo.toml b/prql-compiler/prqlc/Cargo.toml
index a02a8ff1..079471ca 100644
--- a/prql-compiler/prqlc/Cargo.toml
+++ b/prql-compiler/prqlc/Cargo.toml
@@ -20,7 +20,7 @@ concolor-clap = {version = "0.1.0", features = ["api"]}
 env_logger = {version = "0.10.0", features = ["color"]}
 itertools = "0.10.3"
 notify = "^5.1.0"
-prql-compiler = {path = '..', version = "0.8.0"}
+prql-compiler = {path = '..', version = "0.8.1" }
 regex = {version = "1.8.1", features = ["std", "unicode"]}
 serde = "^1"
 serde_json = "1.0.81"
```

ychampion:
## Summary

- Preserve TOML decoration when updating package and workspace versions.
- Preserve inline dependency version formatting when upgrading dependency requirements.

## Why

Fixes #671. The old assignments replaced the whole TOML value, which dropped comments and changed inline-table spacing around updated versions.

## Validation

- `OPENSSL_NO_VENDOR=1 cargo test -p cargo-release --lib --no-default-features --features vendored-libgit2 preserves_ -- --nocapture` - passed
- `OPENSSL_NO_VENDOR=1 cargo test -p cargo-release --lib --no-default-features --features vendored-libgit2 ops::cargo::test -- --nocapture` - passed
- `OPENSSL_NO_VENDOR=1 cargo test -p cargo-release --lib --no-default-features --features vendored-libgit2` - passed
- `OPENSSL_NO_VENDOR=1 cargo clippy -p cargo-release --lib --no-default-features --features vendored-libgit2 -- -D warnings` - passed
- `cargo fmt --check` - passed
- `git diff --check` - passed

### DIFF
--- a/src/ops/cargo.rs
+++ b/src/ops/cargo.rs
@@ -147,7 +147,13 @@ pub fn set_workspace_version(
 ) -> CargoResult<()> {
     let original_manifest = std::fs::read_to_string(manifest_path)?;
     let mut manifest: toml_edit::DocumentMut = original_manifest.parse()?;
-    manifest["workspace"]["package"]["version"] = toml_edit::value(version);
+    overwrite_toml_value(
+        manifest["workspace"]["package"]
+            .as_table_like_mut()
+            .expect("workspace.package table"),
+        "version",
+        version,
+    );
     let manifest = manifest.to_string();
 
     if dry_run {
@@ -249,7 +255,13 @@ pub fn ensure_owners(
 pub fn set_package_version(manifest_path: &Path, version: &str, dry_run: bool) -> CargoResult<()> {
     let original_manifest = std::fs::read_to_string(manifest_path)?;
     let mut manifest: toml_edit::DocumentMut = original_manifest.parse()?;
-    manifest["package"]["version"] = toml_edit::value(version);
+    overwrite_toml_value(
+        manifest["package"]
+            .as_table_like_mut()
+            .expect("package table"),
+        "version",
+        version,
+    );
     let manifest = manifest.to_string();
 
     if dry_run {
@@ -372,7 +384,7 @@ fn upgrade_req(
     version: &semver::Version,
     upgrade: config::DependentVersion,
 ) -> bool {
-    let version_value = if let Some(version_value) = dep_item.get_mut("version") {
+    let version_value = if let Some(version_value) = dep_item.get("version") {
         version_value
     } else {
         log::debug!("not updating path-only dependency on {name}");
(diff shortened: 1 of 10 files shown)


src/ops/cargo.rs
```diff
@@ -541,6 +550,99 @@ mod test {
 
             temp.close().unwrap();
         }
+
+        #[test]
+        fn preserves_version_decor() {
```
epage: Our contrib guide asks for tests to reproduce the undesired behavior in a prior commit. See https://epage.github.io/dev/pr-style/#c-split for more details.

src/ops/cargo.rs
```diff
@@ -541,6 +550,99 @@ mod test {
...
+version = "2.0.0" # release version
+edition = "2021"
+"#,
+            );
+
+            temp.close().unwrap();
+        }
+    }
+
+    mod set_workspace_version {
```
epage: I'd prefer not to do all of this extra nesting

src/ops/cargo.rs
```diff
@@ -541,6 +550,99 @@ mod test {
 
             temp.close().unwrap();
         }
+
+        #[test]
+        fn preserves_version_decor() {
+            let temp = assert_fs::TempDir::new().unwrap();
```
epage: I think these would better fit within our integration tests

src/ops/cargo.rs
```diff
@@ -420,10 +420,19 @@ fn upgrade_req(
         "Updating",
         format!("{manifest_name}'s dependency from {existing_req_str} to {new_req}"),
     );
-    *version_value = toml_edit::value(new_req);
+    set_toml_string(version_value, &new_req);
     true
 }
 
+fn set_toml_string(item: &mut toml_edit::Item, value: &str) {
```
epage: Would it work to instead use the following function from cargo? Would be nice to be consistent.

https://github.com/rust-lang/cargo/blob/f440024b88282844041f4e650f4ed39ff3b62a0b/src/cargo/util/toml_mut/dependency.rs#L676-L687

src/ops/cargo.rs
ychampion: Done. I removed the nested unit-test modules and replaced them with one integration case.

src/ops/cargo.rs
ychampion: Done. The regression now lives under tests/testsuite/version/preserve_version_decor and covers package, workspace-package, and dependency version rewrites.

src/ops/cargo.rs
```diff
@@ -420,10 +432,23 @@ fn upgrade_req(
         "Updating",
         format!("{manifest_name}'s dependency from {existing_req_str} to {new_req}"),
     );
-    *version_value = toml_edit::value(new_req);
+    overwrite_toml_value(dep_item, "version", new_req);
```
epage, line 435: If we're no longer overwriting `version_value`, we don't need it use `get_mut`

src/ops/cargo.rs
epage, line 1: Please squash this where needed. We prefer commit to reflect how a change should be understood for reviewing and then merging, rather than how it was developed.

ychampion: Done. I kept the test-only commit first and folded the immutable lookup follow-up into the fix commit, so the PR is now two commits: test, then implementation. The focused integration test, fmt, and Clippy all pass.

PR #1002, opened 2026-07-31
feat: allow dirty worktrees for publish step

### LINKED ISSUE #1001: feat: allow dirty working trees for the publish step
## Motivation

`cargo release publish` currently rejects a dirty working tree with:

```
error: uncommitted changes detected, please resolve before release
```

This blocks publish-only workflows where generated files, local patches, or other intentional changes are present, even though Cargo itself supports publishing in this state with `cargo publish --allow-dirty`.

Unlike the full `cargo release` workflow, the standalone `publish` step does not bump versions, run pre-release hooks, or create commits. Therefore, allowing a dirty tree here does not risk accidentally committing unrelated changes, which was the concern raised in #854.

## Proposed behavior

Add `--allow-dirty` to `cargo release publish` so that it:

- skips cargo-release's clean-working-tree preflight check for the publish-only step;
- passes `--allow-dirty` through to `cargo publish` during an executing publish;
- preserves the existing clean-tree requirement by default.

This keeps the unsafe full-release behavior unchanged while matching Cargo's explicit opt-in semantics for publish-only workflows.

psandana:
## Summary

- add `--allow-dirty` to `cargo release publish`
- skip the publish step's clean-worktree preflight only when explicitly requested
- forward `--allow-dirty` to executing `cargo publish` commands
- keep the full release workflow clean-only to avoid committing unrelated changes
- document the intentionally narrow scope

## Validation

- `cargo test --lib --bins --all-features`
- `cargo test --test readme --all-features`
- `cargo test --workspace --no-run`
- `cargo fmt --all -- --check`

Closes #1001

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -74,6 +74,10 @@ Push:
 
 ```
 
+The standalone publish step accepts `--allow-dirty` to publish with uncommitted
+changes, matching `cargo publish --allow-dirty`. The full release workflow
+continues to require a clean working tree because it creates commits.
+
 ### Bump level
 
 * `release` (default): Remove the pre-release extension; if any (0.1.0-alpha.1 -> 0.1.0, 0.1.0 -> 0.1.0).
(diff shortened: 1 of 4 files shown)


epage:
Moving to Draft as the Issue isn't accepted yet, per our contrib guide

PR #1005, opened 2026-08-02
fix: check dependencies before publishing

### LINKED ISSUE #624: Verification builds fail fetching (non-dev) dependency that isn't published yet
If I run `cargo release` in the root of: https://github.com/kornelski/cavif-rs commit b2f707ce08bf758048cb967998df92d8dd3d2c7c

https://github.com/kornelski/cavif-rs/tree/b2f707ce08bf758048cb967998df92d8dd3d2c7c/

it chooses not to publish workspace's leaf dependency first (ravif), and then the root crate fails validation (cavif), because its dependency isn't published yet.

```text
$ cargo release
warning: disabled by user, skipping ravif which has files changed since v1.4.0: [
    "./cavif/ravif/Cargo.toml",
    "./cavif/ravif/src/av1encoder.rs",
    "./cavif/ravif/src/dirtyalpha.rs",
    "./cavif/ravif/src/error.rs",
    "./cavif/ravif/src/lib.rs",
]
warning: disabled by user, skipping ravif v0.9.3 despite being unpublished
  Publishing cavif
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   ./cavif/ravif/Cargo.toml
workspace: ./cavif/Cargo.toml
    Updating crates.io index
   Packaging cavif v1.4.1 (./cavif)
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for the requirement `ravif = "^0.9.2"`
  candidate versions found which didn't match: 0.9.1, 0.9.0, 0.8.9, ...
  location searched: crates.io index
  required by package `cavif v1.4.1 (./cavif)`
```

I don't see why `cargo-release` thinks the dependency is "disabled by user". 

Using cargo-release 0.24.0 and rustc 1.65

Guflly:
Checks registry dependencies before publishing and returns an error when the required version is not available. This prevents a release from failing later after other steps have already run.

The check accounts for versions planned in the same release and includes regression coverage for published, unpublished, path-only, git, optional, dev, and workspace dependencies. The full workspace test suite passes locally.

Fixes #624

### DIFF
--- a/src/steps/mod.rs
+++ b/src/steps/mod.rs
@@ -706,3 +706,86 @@ impl BumpLevel {
         Ok(())
     }
 }
+
+pub fn verify_dependencies(
+    selected_pkgs: &[plan::PackageRelease],
+    excluded_pkgs: &[plan::PackageRelease],
+    index: &mut crate::ops::index::CratesIoIndex,
+    dry_run: bool,
+    level: log::Level,
+) -> Result<bool, crate::error::CliError> {
+    let mut success = true;
+
+    for pkg in selected_pkgs {
+        if !pkg.config.publish() || pkg.config.registry().is_some() {
+            continue;
+        }
+        let version = pkg.planned_version.as_ref().unwrap_or(&pkg.initial_version);
+        let mut checked = std::collections::HashSet::new();
+        for dependency in &pkg.meta.dependencies {
+            if dependency.registry.is_some()
+                || (dependency.req == semver::VersionReq::STAR
+                    && !dependency
+                        .source
+                        .as_ref()
+                        .is_some_and(cargo_metadata::Source::is_crates_io))
+                || !checked.insert((dependency.name.as_str(), &dependency.req))
+            {
+                continue;
+            }
+
+            let publishing = selected_pkgs.iter().any(|candidate| {
+                let candidate_version = candidate
+                    .planned_version
+                    .as_ref()
+                    .unwrap_or(&candidate.initial_version);
+                candidate.config.publish()
+                    && candidate.config.registry() == pkg.config.registry()
+                    && candidate.meta.name.as_str() == dependency.name.as_str()
+                    && dependency.req.matches(&candidate_version.full_version)
+            });
+            if publishing
+                || crate::ops::cargo::is_published_req(
+                    index,
+                    pkg.config.registry(),
+                    &dependency.name,
+                    &dependency.req,
+                    pkg.config.certs_source(),
+                )
+            {
+                continue;
+            }
+
+            let workspace_dependency = selected_pkgs
+                .iter()
+                .chain(excluded_pkgs)
+                .find(|candidate| candidate.meta.name.as_str() == dependency.name.as_str());
+            let message = if let Some(workspace_dependency) = workspace_dependency {
+                let dependency_version = workspace_dependency
+                    .planned_version
+                    .as_ref()
+                    .unwrap_or(&workspace_dependency.initial_version);
+                format!(
+                    "{} {} depends on unpublished workspace package {} {}",
+                    pkg.meta.name,
+                    version.full_version_string,
+                    workspace_dependency.meta.name,
+                    dependency_version.full_version_string
+                )
+            } else {
+                format!(
+                    "{} {} depends on unpublished package {} {}",
+                    pkg.meta.name, version.full_version_string, dependency.name, dependency.req
+                )
+            };
+            let _ = crate::ops::shell::log(level, message);
+            success = false;
+        }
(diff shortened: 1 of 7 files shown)


src/steps/mod.rs
```diff
@@ -706,3 +706,86 @@ impl BumpLevel {
...
+        }
+        let version = pkg.planned_version.as_ref().unwrap_or(&pkg.initial_version);
+        let mut checked = std::collections::HashSet::new();
+        for dependency in &pkg.meta.dependencies {
+            if dependency.registry.is_some()
+                || (dependency.req == semver::VersionReq::STAR
+                    && !dependency
+                        .source
+                        .as_ref()
+                        .is_some_and(cargo_metadata::Source::is_crates_io))
```
epage, line 731: Is this sufficient? If you have a git *and* registry dependency, a local `cargo metadata` will just give the git source, right?

tests/testsuite/publish.rs
```diff
@@ -0,0 +1,78 @@
...
+            [workspace]
+            resolver = "2"
+            members = ["dependency", "application"]
+            "#,
+        )
+        .file(
+            "dependency/Cargo.toml",
+            r#"
+            [package]
+            name = "cargo-release-test-unpublished-dependency-624"
```
epage: Let's use limit the package names to just what is meaningful to the test

tests/testsuite/publish.rs
```diff
@@ -0,0 +1,78 @@
...
+    assert!(!output.status.success());
+    let stderr = String::from_utf8(output.stderr).unwrap();
+    assert!(
+        stderr.contains(
+            "cargo-release-test-application-624 0.1.0 depends on unpublished workspace package cargo-release-test-unpublished-dependency-624 0.1.0"
+        ),
+        "{stderr}"
+    );
+    assert!(!stderr.contains("Publishing"), "{stderr}");
+}
```
epage: We should limit ourselves to snapshot tests, like with https://docs.rs/snapbox/latest/snapbox/cmd/struct.OutputAssert.html#method.stdout_eq and `str![]`

tests/testsuite/publish.rs
epage, line 1: Our contrib guide asks for tests to be added in a commit before a fix *with them passing*

See https://epage.github.io/dev/pr-style/#c-test for more details

src/steps/mod.rs
Guflly: It does report the git source, but it still keeps the version requirement. The check only skips git dependencies without a registry version. I added a git + version test for this.

tests/testsuite/publish.rs
Guflly: Updated these to application and dependency.

tests/testsuite/publish.rs
Guflly: Switched both cases to stdout_eq/stderr_eq snapshots.

tests/testsuite/publish.rs
Guflly: Split this into a passing test commit followed by the fix commit.

tests/testsuite/publish.rs
```diff
@@ -0,0 +1,131 @@
...
+        .arg("release")
+        .args(["--package", "application", "--execute", "--no-confirm"])
+        .current_dir(project.root())
+        .assert()
+        .failure()
+        .stdout_eq(str![])
+        .stderr_eq(str![[r#"
+warning: push target `origin/master` doesn't exist
+  Publishing application
+error: crates-io is replaced with remote registry dummy-registry;
```
epage: nit: Since you are already changes, could you update these tests so they get to the point of concern within this test commit so there is a better baseline for comparison of what happened before compared to now?

tests/testsuite/publish.rs
```diff
@@ -0,0 +1,131 @@
...
+include `--registry dummy-registry` or `--registry crates-io`
+
+"#]]);
+}
+
+#[cargo_test]
+fn unpublished_git_dependency() {
```
epage, line 75: We likely also want tests with a git and path dev dependencies which won't have a version and will be stripped on publish

tests/testsuite/publish.rs
Guflly: Updated the pre-fix snapshots to reach Cargo's missing-dependency error, so the fix commit shows the behavior change directly.

tests/testsuite/publish.rs
Guflly: Added unversioned git and path dev dependencies to the fixture. The check ignores both.

epage:
Thanks!

PR #1019, opened 2026-09-16
update libgit2 to pick up relativeworktrees extension

sunshowers:
### What does this PR try to solve?

Without this, `cargo release` fails in a colocated jj repo with multiple workspaces, since that uses relative worktrees under the hood.

LLM involvement: None, just a trivial `cargo update -p libgit2-sys`.


sunshowers:
(The advisories seem like they should be addressed, but aren't made worse by this fix.)

PR #1020, opened 2026-09-17
feat(replace): allow glob pattern in Replace::file

### LINKED ISSUE #104: Allow replacements to apply to multiple files
It would be convenient to have a single `pre-release-replacements` section apply to a number of files in the workspace (eg. for multi-crate repositories). In addition to a `file` key, a `files` key could be accepted that is interpreted as a glob (if it's a string) or a list of globs (if it's a toml array).

FreeMasen:
### What does this PR try to solve?

Closes #104

### Notes to reviewers

In an attempt to address glob patterns in the pre-release-replace configuration this uses the deserializer implementation for that field to convert the input into `Vec<PathBuf>` via the `globwalk` crate.

LLM involvement: none


epage:
Couple of quick notes:
- `globwalk` seems like a dead end crate to use at this point
- we shouldn't be walking the filesystem at config parse time
- if we have a custom deserializer, I'd prefer to use `serde_untagged`
- docs would need updating
- if we could have a test, that would be a good, see also https://epage.github.io/dev/pr-style/#c-test

FreeMasen:
> Couple of quick notes:
> 
>     * `globwalk` seems like a dead end crate to use at this point
> 
>     * we shouldn't be walking the filesystem at config parse time
> 
>     * if we have a custom deserializer, I'd prefer to use `serde_untagged`
> 
>     * docs would need updating
> 
>     * if we could have a test, that would be a good, see also https://epage.github.io/dev/pr-style/#c-test

Thanks for the review and feedback, I'll try and take another pass at this soon

